PR #56, opened 2023-05-13
Use XCURSOR_THEME

mrshiposha:
This PR allows Jay to use `$XCURSOR_THEME` to use non-default xcursor themes.


mahkoh:
Code looks good. Please run `cargo +nightly fmt` so that the rustfmt check succeeds.

mahkoh:
Thanks

PR #59, opened 2023-07-15
fix: xdg_surface configure serial

mrshiposha:
Some applications, like Telegram Desktop, won't send `ack_configure` if the `serial` value in the corresponding `configure` event is 0. Because of that, Telegram's window wasn't showing.

I don't know why Telegram interprets the 0 as some special value of the `serial`.
However, if the `serial` starts with 1, Telegram suddenly works (i.e., it shows the window).

### DIFF
--- a/src/ifs/wl_surface/xdg_surface.rs
+++ b/src/ifs/wl_surface/xdg_surface.rs
@@ -347,7 +347,7 @@ impl SurfaceExt for XdgSurface {
                     if let Some(ext) = self.ext.get() {
                         ext.initial_configure()?;
                     }
-                    self.send_configure(rse + 1);
+                    self.send_configure(rse);
                 }
                 // return CommitAction::AbortCommit;
             }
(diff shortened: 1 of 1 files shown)


mahkoh:
I think it makes more sense to do

```rust
requested_serial: NumCell::new(1),
```

in line 99.

>I don't know why Telegram interprets the 0 as some special value of the serial.

I assume that it interprets 0 as "I have never received a configure event".

mahkoh:
Thanks

davidedmundson:
>I assume that it interprets 0 as "I have never received a configure event".

~ish.

There's a path to avoid acking the same thing twice, that compares against the last acked value which defaults to zero. 
If a compositor uses libwayland's wl_display_next_serial for the serial we always start on 1 like this patch does.

mrshiposha:
>I assume that it interprets 0 as "I have never received a configure event".

I could suggest that Telegram (or, more precisely, the particular version of Qt it uses, since other Qt apps like VLC work properly) uses a `uint` initialized to 0 to store serials. When it sees `0`, it thinks it is not a new serial (since it is not greater than the "old" serial, which is 0) and just ignores it.
At the same time, other apps use `int` initialized to -1

PR #177, opened 2024-04-21
wayland: implement wl_touch

### LINKED ISSUE #115: Feature: wl_touch support
I don't have any desktop devices with a touchscreen so this is unlikely to make progress.

Sporif:
I'm testing this on a touchscreen laptop and trying to implement basic support:

- [x] Send events to the active client at the correct position (tested by tapping, scrolling and zooming in Konsole and Firefox)
- [x] Hide the cursor on touch down
- [x] Switch the active client on touch down
- [x] Switch client on titlebar touch down
- [x] Switch workspace on bar touch down
- [x] Resize clients with touch motion on the edges/corners

Future work:
- calibration-matrix option for switching the orientation of a touchscreen
- touch gestures/bindings

Closes #115

### DIFF
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -232,6 +232,23 @@ pub enum KeyState {
 }
 
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub struct TouchPosition {
+    pub x: Fixed,
+    pub y: Fixed,
+    pub x_transformed: Fixed,
+    pub y_transformed: Fixed,
+}
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum TouchEvent {
+    Down { pos: TouchPosition },
+    Up,
+    Motion { pos: TouchPosition },
+    Cancel,
+    Frame,
+}
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
 pub enum ScrollAxis {
     Horizontal = HORIZONTAL_SCROLL as _,
     Vertical = VERTICAL_SCROLL as _,
@@ -270,6 +287,12 @@ pub enum InputEvent {
         time_usec: u64,
         button: u32,
         state: KeyState,
+    },
+
+    Touch {
+        seat_slot: i32,
+        time_usec: u64,
+        event: TouchEvent,
     },
 
     AxisPx {
--- a/src/backends/metal/input.rs
+++ b/src/backends/metal/input.rs
@@ -1,6 +1,6 @@
 use {
     crate::{
-        backend::{AxisSource, InputEvent, KeyState, ScrollAxis},
+        backend::{AxisSource, InputEvent, KeyState, ScrollAxis, TouchEvent, TouchPosition},
         backends::metal::MetalBackend,
         fixed::Fixed,
         ifs::wl_seat::tablet::{
@@ -121,6 +121,11 @@ impl MetalBackend {
             c::LIBINPUT_EVENT_TABLET_PAD_BUTTON => self.handle_tablet_pad_button(event),
             c::LIBINPUT_EVENT_TABLET_PAD_RING => self.handle_tablet_pad_ring(event),
             c::LIBINPUT_EVENT_TABLET_PAD_STRIP => self.handle_tablet_pad_strip(event),
+            c::LIBINPUT_EVENT_TOUCH_DOWN => self.handle_touch_down(event),
+            c::LIBINPUT_EVENT_TOUCH_UP => self.handle_touch_up(event),
+            c::LIBINPUT_EVENT_TOUCH_MOTION => self.handle_touch_motion(event),
+            c::LIBINPUT_EVENT_TOUCH_CANCEL => self.handle_touch_cancel(event),
+            c::LIBINPUT_EVENT_TOUCH_FRAME => self.handle_touch_frame(event),
             _ => {}
         }
     }
@@ -544,5 +549,62 @@ impl MetalBackend {
                 n => Some(n),
             },
         });
+    }
+
+    fn handle_touch_down(self: &Rc<Self>, event: LibInputEvent) {
+        let (event, dev) = unpack!(self, event, touch_event);
+        let pos = TouchPosition {
+            x: Fixed::from_f64(event.x()),
+            y: Fixed::from_f64(event.y()),
+            x_transformed: Fixed::from_f64(event.x_transformed(1)),
+            y_transformed: Fixed::from_f64(event.y_transformed(1)),
+        };
+        dev.event(InputEvent::Touch {
+            seat_slot: event.seat_slot(),
+            time_usec: event.time_usec(),
+            event: TouchEvent::Down { pos },
+        })
+    }
+
+    fn handle_touch_up(self: &Rc<Self>, event: LibInputEvent) {
+        let (event, dev) = unpack!(self, event, touch_event);
+        dev.event(InputEvent::Touch {
+            seat_slot: event.seat_slot(),
+            time_usec: event.time_usec(),
+            event: TouchEvent::Up,
+        })
+    }
+
+    fn handle_touch_motion(self: &Rc<Self>, event: LibInputEvent) {
+        let (event, dev) = unpack!(self, event, touch_event);
+        let pos = TouchPosition {
+            x: Fixed::from_f64(event.x()),
+            y: Fixed::from_f64(event.y()),
+            x_transformed: Fixed::from_f64(event.x_transformed(1)),
+            y_transformed: Fixed::from_f64(event.y_transformed(1)),
+        };
+        dev.event(InputEvent::Touch {
+            seat_slot: event.seat_slot(),
+            time_usec: event.time_usec(),
+            event: TouchEvent::Motion { pos },
+        })
+    }
+
+    fn handle_touch_cancel(self: &Rc<Self>, event: LibInputEvent) {
+        let (event, dev) = unpack!(self, event, touch_event);
+        dev.event(InputEvent::Touch {
+            seat_slot: event.seat_slot(),
+            time_usec: event.time_usec(),
+            event: TouchEvent::Cancel,
+        })
+    }
+
+    fn handle_touch_frame(self: &Rc<Self>, event: LibInputEvent) {
+        let (event, dev) = unpack!(self, event, touch_event);
+        dev.event(InputEvent::Touch {
+            seat_slot: 0,
+            time_usec: event.time_usec(),
+            event: TouchEvent::Frame,
+        })
     }
 }
--- a/src/cli/seat_test.rs
+++ b/src/cli/seat_test.rs
@@ -15,7 +15,8 @@ use {
                 TabletPadStripSource, TabletPadStripStop, TabletToolButton, TabletToolDistance,
                 TabletToolDown, TabletToolFrame, TabletToolMotion, TabletToolPressure,
                 TabletToolProximityIn, TabletToolProximityOut, TabletToolRotation,
-                TabletToolSlider, TabletToolTilt, TabletToolUp, TabletToolWheel,
+                TabletToolSlider, TabletToolTilt, TabletToolUp, TabletToolWheel, TouchCancel,
+                TouchDown, TouchFrame, TouchMotion, TouchUp,
             },
         },
     },
@@ -582,6 +583,63 @@ async fn run(seat_test: Rc<SeatTest>) {
             print!(", stop");
         }
         println!();
+    });
+    let st = seat_test.clone();
+    TouchDown::handle(tc, se, (), move |_, ev| {
+        if all || ev.seat == seat {
+            if all {
+                print!("Seat: {}, ", st.name(ev.seat));
+            }
+            println!(
+                "Time: {:.4}, Touch: {}, Down: {}x{}",
+                time(ev.time_usec),
+                ev.id,
+                ev.x,
+                ev.y
+            );
+        }
+    });
+    let st = seat_test.clone();
+    TouchUp::handle(tc, se, (), move |_, ev| {
+        if all || ev.seat == seat {
+            if all {
+                print!("Seat: {}, ", st.name(ev.seat));
+            }
+            println!("Time: {:.4}, Touch: {}, Up", time(ev.time_usec), ev.id);
+        }
+    });
+    let st = seat_test.clone();
+    TouchMotion::handle(tc, se, (), move |_, ev| {
+        if all || ev.seat == seat {
+            if all {
+                print!("Seat: {}, ", st.name(ev.seat));
+            }
+            println!(
+                "Time: {:.4}, Touch: {} Motion: {}x{}",
+                time(ev.time_usec),
+                ev.id,
+                ev.x,
+                ev.y
+            );
+        }
+    });
+    let st = seat_test.clone();
+    TouchFrame::handle(tc, se, (), move |_, ev| {
+        if all || ev.seat == seat {
+            if all {
+                print!("Seat: {}, ", st.name(ev.seat));
+            }
+            println!("Time: {:.4}, Touch: {}, Frame", time(ev.time_usec), ev.id);
+        }
+    });
+    let st = seat_test.clone();
+    TouchCancel::handle(tc, se, (), move |_, ev| {
+        if all || ev.seat == seat {
+            if all {
+                print!("Seat: {}, ", st.name(ev.seat));
+            }
+            println!("Time: {:.4}, Touch: {}, Cancel", time(ev.time_usec), ev.id);
+        }
     });
     pending::<()>().await;
 }
--- a/src/ifs/wl_seat.rs
+++ b/src/ifs/wl_seat.rs
@@ -6,6 +6,7 @@ mod kb_owner;
 mod pointer_owner;
 pub mod tablet;
 pub mod text_input;
+mod touch_owner;
 pub mod wl_keyboard;
 pub mod wl_pointer;
 pub mod wl_touch;
@@ -22,6 +23,7 @@ pub mod zwp_virtual_keyboard_v1;
 use {
     crate::{
         async_engine::SpawnedFuture,
+        backend::InputDeviceCapability,
         client::{Client, ClientError, ClientId},
         cursor_user::{CursorUser, CursorUserGroup, CursorUserOwner},
         fixed::Fixed,
@@ -52,6 +54,7 @@ use {
                     zwp_input_method_keyboard_grab_v2::ZwpInputMethodKeyboardGrabV2,
                     zwp_input_method_v2::ZwpInputMethodV2, zwp_text_input_v3::ZwpTextInputV3,
                 },
+                touch_owner::TouchOwnerHolder,
                 wl_keyboard::{WlKeyboard, WlKeyboardError, REPEAT_INFO_SINCE},
                 wl_pointer::WlPointer,
                 wl_touch::WlTouch,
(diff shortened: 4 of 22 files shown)


mahkoh:
Thanks for working on this.

I'm not familiar with how other compositors implement touch input, but here are some thoughts regarding how I would imagine it:

There are two touch owners:

```rust
struct DefaultTouchOwner;

struct GrabTouchOwner {
    node: Rc<dyn Node>,
    down_ids: AHashSet<i32>,
}
```

The DefaultTouchOwner ignores all events except `down`. Once it gets a `down` event, it finds the node under the touch position and replaces itself by GrabTouchOwner.

The down_ids field contains all ids that are currently down. When this set becomes empty, the GrabTouchOwner replaces itself by the DefaultTouchOwner.

When the GrabTouchOwner receives a `down` event, that event is always sent to the `node`, regardless of where on the screen it occurs. This means that you can press with one hand on an application and then with another hand anywhere else on the screen and the second event is sent to the node that was hit by the first press.

-------

The first down event should also switch the keyboard focus to that node if it is an application window.

mahkoh:
Alternatively, each touch id could have its touch owner. This seems to be how it works on android.

Sporif:
I've implemented your method, what do you think?

Also, I was trying to hide the cursor with `seat.set_app_cursor(None) `, but then it doesn't show up again until it enters another surface. What's the best way of implementing that feature?

src/ifs/wl_seat/event_handling.rs
```diff
@@ -234,6 +241,34 @@ impl WlSeatGlobal {
         }
     }
 
+    fn touch_event(self: &Rc<Self>, id: i32, time_usec: u64, event: TouchEvent) {
+        let pos_transformed = |pos: TouchPosition| {
+            let output_pos = self.output.get().global.pos.get();
+            let x = Fixed::from_f64(pos.x_transformed.to_f64() * f64::from(output_pos.width()));
+            let y = Fixed::from_f64(pos.y_transformed.to_f64() * f64::from(output_pos.height()));
+            (x, y)
+        };
```
mahkoh: `self.output` is the output that currently contains the pointer. If there are multiple outputs, then this code is not going to work. For any touch input device, touch input should always map to a specific output. There will need to be a new configuration option to map input devices to outputs.

src/backends/metal/input.rs
```diff
@@ -91,6 +91,11 @@ impl MetalBackend {
             c::LIBINPUT_EVENT_POINTER_SCROLL_CONTINUOUS => {
                 self.handle_pointer_axis(event, AxisSource::Continuous)
             }
+            c::LIBINPUT_EVENT_TOUCH_DOWN => self.handle_touch_down(event),
+            c::LIBINPUT_EVENT_TOUCH_UP => self.handle_touch_up(event),
+            c::LIBINPUT_EVENT_TOUCH_MOTION => self.handle_touch_motion(event),
+            c::LIBINPUT_EVENT_TOUCH_CANCEL => self.handle_touch_cancel(event),
+            c::LIBINPUT_EVENT_TOUCH_FRAME => self.handle_touch_frame(event),
```
mahkoh, line 128: Wayland also has shape and orientation events. Are these not supported by libinput?

src/ifs/wl_seat.rs
```diff
@@ -1202,7 +1212,7 @@ impl WlSeat {
     fn send_capabilities(self: &Rc<Self>) {
         self.client.event(Capabilities {
             self_id: self.id,
-            capabilities: POINTER | KEYBOARD,
+            capabilities: POINTER | KEYBOARD | TOUCH,
```
mahkoh: We might want to only advertise touch if any touch devices are connected.

src/ifs/wl_seat/touch_owner.rs
```diff
@@ -0,0 +1,139 @@
...
+        let y_int = y.round_down();
+        found_tree.push(FoundNode {
+            node: seat.state.root.clone(),
+            x: x_int,
+            y: y_int,
+        });
+        seat.state
+            .root
+            .node_find_tree_at(x_int, y_int, &mut found_tree, FindTreeUsecase::None);
+        if let Some(node) = found_tree.last() {
```
mahkoh, line 119: Touch focus has to be added to `NodeSeatState`.

src/ifs/wl_seat/touch_owner.rs
```diff
@@ -0,0 +1,139 @@
...
+
+    fn cancel(&self, _seat: &Rc<WlSeatGlobal>) {
+        // nothing
+    }
+}
+
+impl TouchOwner for GrabTouchOwner {
+    fn down(&self, seat: &Rc<WlSeatGlobal>, time_usec: u64, id: i32, x: Fixed, y: Fixed) {
+        self.down_ids.borrow_mut().insert(id);
+        self.node.node_on_touch_down(seat, time_usec, id, x, y);
```
mahkoh: `x` and `y` should be relative to the node position.

src/ifs/wl_seat/touch_owner.rs
```diff
@@ -0,0 +1,139 @@
...
+        self.node.node_on_touch_motion(seat, time_usec, id, x, y);
+    }
+
+    fn frame(&self, seat: &Rc<WlSeatGlobal>) {
+        self.node.node_on_touch_frame(seat);
+    }
+
+    fn cancel(&self, seat: &Rc<WlSeatGlobal>) {
+        self.node.node_on_touch_cancel(seat);
+    }
```
mahkoh: This should revert to the default owner.

src/ifs/wl_seat/touch_owner.rs
```diff
@@ -0,0 +1,139 @@
...
+            let grab_owner = Rc::new(GrabTouchOwner {
+                node: node.node.clone(),
+                down_ids: RefCell::new(AHashSet::new()),
+            });
+            grab_owner.down(seat, time_usec, id, x, y);
+            if let Some(surface) = node.node.clone().node_into_surface() {
+                if let Some(node) = surface.get_focus_node(seat.id) {
+                    seat.focus_node(node);
+                }
+            }
```
mahkoh: This should be done in the callback. E.g. `touch_down_surface`.

src/backends/metal/input.rs
Sporif: It doesn't look like it: https://gitlab.freedesktop.org/libinput/libinput/-/issues/746

Sporif:
> LGTM. Have you checked how other compositors handle multi-touch?

I haven't looked at the code but KWin seems to do it the way you suggested. Sway only sends touch down events that fall within the boundaries of the initial surface after the first touch down, but it always sends motion events.

> Why do you want to hide the cursor?

KWin, Sway and Windows all hide the cursor on touch. I find it can be distracting when you're not using it.

src/ifs/wl_seat/touch_owner.rs
Sporif: I'm not sure what to do here.  Add `touch_foci` and create new `touch_focus/touch_unfocus` functions? Or add it to `focus/unfocus`?

src/ifs/wl_seat.rs
Sporif: It looks like the capabilities event is only sent to clients once at the start, shouldn't be sent every time an input device is added/removed?

mahkoh: Currently the capabilities are always `POINTER | KEYBOARD` so there is no need for this. But if the capabilities can actually change then this event should be sent whenever the capabilities change.

src/ifs/wl_seat/touch_owner.rs
mahkoh: `touch_foci` is what I had in mind.

mahkoh:
>KWin, Sway and Windows all hide the cursor on touch. I find it can be distracting when you're not using it.

Maybe it's best to simply not render the cursor in that case. See `GfxFramebuffer::create_render_pass`.

src/backends/metal/input.rs
Sporif: Since these events are not supported I think this is resolved.

src/ifs/wl_seat.rs
Sporif: I added a check for any connected touch devices.

src/ifs/wl_seat/event_handling.rs
Sporif: I added a config option `mapped-output` and a fallback to the built-in output, if any.

mahkoh:
I'm currently working on adding tablet support. I'll continue the review once that is done.

mahkoh:
The tablet PR #190 has significant overlap with this MR:

- It adds a mechanism to map input devices to outputs.
- It adds a CursorUserGroup type with a deactivate function that hides the cursor.
- It adds the necessary code for tablets to interact with compositor GUI elements.

Sporif:
> The tablet PR #190 has significant overlap with this MR:
> 
> * It adds a mechanism to map input devices to outputs.
> * It adds a CursorUserGroup type with a deactivate function that hides the cursor.
> * It adds the necessary code for tablets to interact with compositor GUI elements.

So should I wait for you to merge that PR and then rework this one?

mahkoh:
#190 has been merged.

Sporif:
I've now implemented all the basic features. There's still some more future work but for now I think this is enough.

src/backend.rs
```diff
@@ -231,6 +231,23 @@ pub enum KeyState {
...
+}
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum TouchEvent {
+    Down { pos: TouchPosition },
+    Up,
+    Motion { pos: TouchPosition },
+    Cancel,
+    Frame,
+}
```
mahkoh, line 249: This should be flattened into the InputEvent enum and the fields of TouchPosition should be flattened into the Down and Motion variants.

src/backend.rs
```diff
@@ -272,6 +289,12 @@ pub enum InputEvent {
         state: KeyState,
     },
 
+    Touch {
+        seat_slot: i32,
```
mahkoh, line 293: This should be named something else due to the existing meaning of seat.

src/cli/seat_test.rs
```diff
@@ -583,6 +584,63 @@ async fn run(seat_test: Rc<SeatTest>) {
         }
         println!();
     });
+    let st = seat_test.clone();
+    TouchDown::handle(tc, se, (), move |_, ev| {
+        if all || ev.seat == seat {
```
mahkoh, line 589: The events should be collected and only be printed on the next frame event. Similar to the tablet frame event.

src/ifs/wl_seat.rs
```diff
@@ -883,13 +891,24 @@ impl WlSeatGlobal {
...
     ) -> Result<(), WlSeatError> {
+        let mut capabilities = POINTER | KEYBOARD;
+        let handlers = &self.state.input_device_handlers;
+        for (_, d) in handlers.borrow().iter() {
+            let dev = &d.data.device;
+            if dev.has_capability(InputDeviceCapability::Touch) {
+                capabilities |= TOUCH;
+                break;
+            }
+        }
```
mahkoh, line 902: The capabilities of a seat should only consider the input devices connected to a seat. These capabilities can also change at runtime when a device is attached or detached and then they will have to be sent to the client again.

src/ifs/wl_seat.rs
```diff
@@ -1053,10 +1074,23 @@ impl WlSeatRequestHandler for WlSeat {
...
-        Ok(())
+        if self.capabilities & TOUCH == 0 {
+            self.client.protocol_error(
+                self,
+                MISSING_CAPABILITY,
+                &format!(
+                    "wl_seat {} .get_touch called when no touch capability has existed",
+                    self.id
+                ),
+            );
```
mahkoh, line 1085: This can be removed.

src/ifs/wl_seat/event_handling.rs
```diff
@@ -278,6 +292,7 @@ impl WlSeatGlobal {
             InputEvent::TabletPadModeSwitch { .. } => {}
             InputEvent::TabletPadRing { .. } => {}
             InputEvent::TabletPadStrip { .. } => {}
+            InputEvent::Touch { .. } => {}
```
mahkoh, line 295: This should call `self.cursor_group.deactivate`.

src/ifs/wl_seat/event_handling.rs
```diff
@@ -599,6 +619,52 @@ impl WlSeatGlobal {
...
+                    .get()
+                    .and_then(|out| out.node.get())
+                    .or_else(|| {
+                        self.state
+                            .builtin_output
+                            .get()
+                            .and_then(|con| self.state.root.outputs.get(&con))
+                    })
+                    .map(|o| o as Rc<dyn Node>)
+                    .unwrap_or_else(|| self.state.root.clone());
```
mahkoh, line 645: This should use `DeviceHandlerData::get_rect`.

src/ifs/wl_seat/touch_owner.rs
```diff
@@ -0,0 +1,197 @@
...
+        seat: &Rc<WlSeatGlobal>,
+        mapped_node: Rc<dyn Node>,
+        time_usec: u64,
+        id: i32,
+        x: Fixed,
+        y: Fixed,
+    ) {
+        seat.cursor_group().deactivate();
+        let pos = mapped_node.node_absolute_position();
+        let (x, y) = transform_abs(x, y, pos);
```
mahkoh, line 109: This should be done outside of touch_owner.rs.

src/ifs/wl_seat/touch_owner.rs
```diff
@@ -0,0 +1,197 @@
...
+            node.node_seat_state().touch_begin(seat);
+            let down_ids = RefCell::new(AHashSet::new());
+            down_ids.borrow_mut().insert(id);
+            let (x_rel, y_rel) = transform_rel(x, y, node.node_absolute_position());
+            seat.touch_owner.owner.set(Rc::new(GrabTouchOwner {
+                pos,
+                node: node.clone(),
+                down_ids,
+            }));
+            node.node_on_touch_down(seat, time_usec, id, x_rel, y_rel);
```
mahkoh, line 130: After setting the new owner, this should just call `grab_owner.down(...)`.

src/ifs/wl_seat/touch_owner.rs
```diff
@@ -0,0 +1,197 @@
...
+    fn down(
+        &self,
+        seat: &Rc<WlSeatGlobal>,
+        _mapped_node: Rc<dyn Node>,
+        time_usec: u64,
+        id: i32,
+        x: Fixed,
+        y: Fixed,
+    ) {
+        self.down_ids.borrow_mut().insert(id);
```
mahkoh, line 162: If the `id` is already down, this should return immediately.

src/ifs/wl_seat/touch_owner.rs
```diff
@@ -0,0 +1,197 @@
...
+        let (x, y) = transform_abs(x, y, self.pos);
+        let (x_rel, y_rel) = transform_rel(x, y, self.node.node_absolute_position());
+        self.node
+            .clone()
+            .node_on_touch_down(seat, time_usec, id, x_rel, y_rel);
+    }
+
+    fn up(&self, seat: &Rc<WlSeatGlobal>, time_usec: u64, id: i32) {
+        self.down_ids.borrow_mut().remove(&id);
```
mahkoh, line 171: If the `id` is already up, this should return immediately.

src/ifs/wl_seat/touch_owner.rs
```diff
@@ -0,0 +1,197 @@
...
+        self.node.clone().node_on_touch_up(seat, time_usec, id);
+        if self.down_ids.borrow().is_empty() {
+            self.node.node_seat_state().touch_end(seat);
+            seat.touch_owner.clear();
+        }
+    }
+
+    fn motion(&self, seat: &Rc<WlSeatGlobal>, time_usec: u64, id: i32, x: Fixed, y: Fixed) {
+        self.down_ids.borrow_mut().insert(id);
```
mahkoh, line 180: This should be removed.

src/libinput/event.rs
```diff
@@ -467,3 +480,28 @@ impl<'a> LibInputEventTabletPad<'a> {
...
+    pub fn seat_slot(&self) -> i32 {
+        unsafe { libinput_event_touch_get_seat_slot(self.event) }
+    }
+    pub fn x(&self) -> f64 {
+        unsafe { libinput_event_touch_get_x(self.event) }
+    }
+
+    pub fn y(&self) -> f64 {
+        unsafe { libinput_event_touch_get_y(self.event) }
+    }
```
mahkoh, line 494: I don't think we need these.

src/tree.rs
```diff
@@ -200,6 +200,50 @@ pub trait Node: 'static {
...
+        let _ = seat;
+        let _ = time_usec;
+        let _ = id;
+        let _ = x;
+        let _ = y;
+    }
+
+    fn node_on_touch_frame(&self, seat: &WlSeatGlobal) {
+        let _ = seat;
+    }
```
mahkoh, line 241: I think this should work more like `node_on_axis_event` where the contents of a frame are collected first and then passed via a single call.

src/ifs/wl_seat/touch_owner.rs
```diff
@@ -0,0 +1,197 @@
...
+impl Default for TouchOwnerHolder {
+    fn default() -> Self {
+        Self {
+            default: Rc::new(DefaultTouchOwner),
+            owner: CloneCell::new(Rc::new(DefaultTouchOwner)),
+        }
+    }
+}
+
+impl TouchOwnerHolder {
```
mahkoh, line 27: This also needs to handle the case where the node that we're interacting with changes its position. See `WlSeatGlobal::apply_changes`.

src/tree/container.rs
```diff
@@ -1275,6 +1276,34 @@ impl Node for ContainerNode {
...
+
+    fn node_on_touch_motion(
+        self: Rc<Self>,
+        seat: &WlSeatGlobal,
+        _time_usec: u64,
+        id: i32,
+        x: Fixed,
+        y: Fixed,
+    ) {
+        self.pointer_move(CursorType::Touch(id), seat.pointer_cursor(), x, y, false);
```
mahkoh, line 1304: `pointer_move` should be refactor so that the cursor is optional.

mahkoh (changes requested):

src/ifs/wl_seat/touch_owner.rs
Sporif: What changes should be applied in such a case? There's no touch cursor, and for the duration of the touch sequence the node should stay the same no matter where it is. 

Edit: And the latest node position is used for every touch down or motion.

src/ifs/wl_seat/event_handling.rs
Sporif: I want it to fallback to `state.builtin_output` if there's no mapped output, but `get_rect` always returns a position. And the node is required anyways.

src/ifs/wl_seat/event_handling.rs
Sporif: Should it be called for all Touch events? I think at least a touch cancel shouldn't.

src/ifs/wl_seat/event_handling.rs
mahkoh: I believe this should work identical to tablets. If there is a way to match an input device to an output device (possibly via the syspath, but this would have to be carefully analyzed), then this could serve as a default.

>And the node is required anyways.

I don't believe it is. See the equivalent code for tablets.

src/ifs/wl_seat/event_handling.rs
mahkoh: Touch down and motion should be sufficient.

src/ifs/wl_seat/touch_owner.rs
mahkoh: I think we'll have to store the current absolute positions of the touch ids in compositor space and then send synthetic motion events to the node when the node moves.

mahkoh:
I've opened #230 to fix the remaining issues with this PR.

One thing I've removed is integration with the compositor UI. There were some serious issues with multi-touch and I'm only testing this with a graphics tablet hacked to work as a touch screen, so I don't have the fine control to properly design this. This can get added back in the future.

PR #218, opened 2024-06-28
pointer movement during focus changes

junelva:
These commits make two changes: the addition of the command `"center-pointer-on-focused"` and the addition of the option `move-pointer` (bool) to the `"show-workspace"` command type.

This was needed for my two-monitor setup. In my practice, it looks like this in the config:

```toml
# example
# alt-space = "center-pointer-on-focused"


alt-h = ["focus-left", "center-pointer-on-focused"]
alt-j = ["focus-down", "center-pointer-on-focused"]
alt-k = ["focus-up", "center-pointer-on-focused"]
alt-l = ["focus-right", "center-pointer-on-focused"]
```
and
```toml

alt-1 = { type = "show-workspace", name = "1", move-pointer = true }
alt-2 = { type = "show-workspace", name = "2", move-pointer = true }
alt-3 = { type = "show-workspace", name = "3", move-pointer = true }
alt-4 = { type = "show-workspace", name = "4", move-pointer = true }
alt-5 = { type = "show-workspace", name = "5", move-pointer = true }
alt-6 = { type = "show-workspace", name = "6", move-pointer = true }
alt-7 = { type = "show-workspace", name = "7", move-pointer = true }
alt-8 = { type = "show-workspace", name = "8", move-pointer = true }
alt-9 = { type = "show-workspace", name = "9", move-pointer = true }
alt-0 = { type = "show-workspace", name = "10", move-pointer = true }
```

A known issue is that pairing `"center-pointer-on-focused"` with a command such as `"move-left"`, the cursor actually snaps to the wrong location. Perhaps a bug in move (tl rect not updated?) but I didn't look into it yet. Another issue is that when alacritty hides the cursor during typing (a feature that I like!), the activation that was necessary in order to move cross-monitor can sometimes prevent that behavior from resuming - but it will after toggling fullscreen, so perhaps activating the cursor before the cursor jump (if invisible), then deactivating it again to the previous state will be necessary.

Additional work I intend to do is allowing the `"focus-*"` commands to move between outputs/monitors. But I wanted to submit the PR for review in case I'm doing this all wrong. While it's working for me locally already, there may be other issues. If you get a chance to look at it, I'm interested in your thoughts.

One contribution question I had - I wasn't sure about updating `spec.yaml`. I did so for the option, but not the new command. Should that be completed also?

Edit: Looks like the test for `"show-workspace"` will need to be updated too.

### DIFF
--- a/jay-config/src/_private/ipc.rs
+++ b/jay-config/src/_private/ipc.rs
@@ -341,6 +341,9 @@ pub enum ClientMessage<'a> {
     DisablePointerConstraint {
         seat: Seat,
     },
+    CenterPointerOnFocused {
+        seat: Seat,
+    },
     ConnectorSetEnabled {
         connector: Connector,
         enabled: bool,
--- a/jay-config/src/_private/ipc.rs
+++ b/jay-config/src/_private/ipc.rs
@@ -256,6 +256,7 @@ pub enum ClientMessage<'a> {
     ShowWorkspace {
         seat: Seat,
         workspace: Workspace,
+        move_pointer: bool,
     },
     SetWorkspace {
         seat: Seat,
--- a/jay-config/src/input.rs
+++ b/jay-config/src/input.rs
@@ -356,8 +356,8 @@ impl Seat {
     ///
     /// If the workspace doesn't currently exist, it is created on the output that contains the
     /// seat's cursor.
-    pub fn show_workspace(self, workspace: Workspace) {
-        get!().show_workspace(self, workspace)
+    pub fn show_workspace(self, workspace: Workspace, move_pointer: bool) {
+        get!().show_workspace(self, workspace, move_pointer)
     }
 
     /// Moves the currently focused window to the workspace.
--- a/jay-config/src/input.rs
+++ b/jay-config/src/input.rs
@@ -385,6 +385,11 @@ impl Seat {
         get!().disable_pointer_constraint(self)
     }
 
+    /// Centers the mouse pointer on the focused window.
+    pub fn center_pointer_on_focused(self) {
+        get!().center_pointer_on_focused(self)
+    }
+
     /// Moves the currently focused workspace to another output.
     pub fn move_to_output(self, connector: Connector) {
         get!().move_to_output(WorkspaceSource::Seat(self), connector);
--- a/src/ifs/wl_seat.rs
+++ b/src/ifs/wl_seat.rs
@@ -405,6 +405,12 @@ impl WlSeatGlobal {
         }
     }
 
+    pub fn center_pointer_on_focused(&self) {
+        let kb_node = self.keyboard_node.get();
+        let (x, y) = kb_node.node_absolute_position().center();
+        self.pointer_cursor.set_position(Fixed(x), Fixed(y));
+    }
+
     fn maybe_constrain_pointer_node(&self) {
         if let Some(pn) = self.pointer_node() {
             if let Some(surface) = pn.node_into_surface() {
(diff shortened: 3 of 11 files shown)


jay-config/src/_private/ipc.rs
```diff
@@ -256,6 +256,7 @@ pub enum ClientMessage<'a> {
     ShowWorkspace {
         seat: Seat,
         workspace: Workspace,
+        move_pointer: bool,
     },
```
mahkoh, line 260: You cannot modify messages because that would break existing config.sos. See https://github.com/mahkoh/jay/blob/f94f199ab12a4545a455a04aec11ec84b62f7c27/jay-config/src/_private/client.rs#L1053-L1066 for how to handle this.

jay-config/src/_private/ipc.rs
```diff
@@ -341,6 +342,9 @@ pub enum ClientMessage<'a> {
     DisablePointerConstraint {
         seat: Seat,
     },
+    CenterPointerOnFocused {
+        seat: Seat,
+    },
```
mahkoh, line 347: New messages must be added at the bottom for the same reason.

jay-config/src/input.rs
```diff
@@ -356,8 +356,8 @@ impl Seat {
     ///
     /// If the workspace doesn't currently exist, it is created on the output that contains the
     /// seat's cursor.
-    pub fn show_workspace(self, workspace: Workspace) {
-        get!().show_workspace(self, workspace)
+    pub fn show_workspace(self, workspace: Workspace, move_pointer: bool) {
```
mahkoh, line 359: This would break the API. You have to add a new function instead.

toml-config/src/config/parsers/action.rs
```diff
@@ -143,7 +144,11 @@ impl ActionParser<'_> {
 
     fn parse_show_workspace(&mut self, ext: &mut Extractor<'_>) -> ParseResult<Self> {
         let name = ext.extract(str("name"))?.value.to_string();
-        Ok(Action::ShowWorkspace { name })
+        let move_pointer = match ext.extract(opt(bol("move-pointer")))? {
```
mahkoh, line 147: You should only call `extract` once. See other places in this file for how to extract multiple fields at once.

toml-spec/spec/spec.yaml
```diff
@@ -204,6 +204,13 @@ Action:
               description: The name of the workspace.
```
mahkoh, line 204: You also have to document center-pointer-on-focused

src/ifs/wl_seat.rs
```diff
@@ -405,6 +405,15 @@ impl WlSeatGlobal {
         }
     }
 
+    pub fn center_pointer_on_focused(&self) {
+        let kb_node = self.keyboard_node.get();
+        if let Some(tl) = kb_node.node_toplevel() {
+            let (x, y) = tl.node_absolute_position().center();
+            self.pointer_cursor
+                .set_position(Fixed::from_int(x), Fixed::from_int(y));
```
mahkoh, line 413: You have to call `self.cursor_moved(time_usec)` after this.

mahkoh:
>A known issue is that pairing "center-pointer-on-focused" with a command such as "move-left", the cursor actually snaps to the wrong location.

This is because the new layout is only scheduled to be computed at this point but has not actually been computed yet: https://github.com/mahkoh/jay/blob/f94f199ab12a4545a455a04aec11ec84b62f7c27/src/tree/container.rs#L949

>Another issue is that when alacritty hides the cursor during typing (a feature that I like!), the activation that was necessary in order to move cross-monitor can sometimes prevent that behavior from resuming

This is because you don't call `cursor_moved` so the cursor owner is not updated.

mahkoh (changes requested):

junelva:
I really appreciate the feedback. It's very helpful. For now, I'm not going to continue work on this.

PR #223, opened 2024-07-10
Implement damage tracking

### LINKED ISSUE #28: Feature: Damage tracking
Moved to post 1.0 as a non-essential enhancement.

mahkoh:
Close #28


cwegener:
Nice work.
I gave it a quick test on my old i7-7500U with HD620 iGPU.
The resulting CPU usage of damage tracking is VERY high, especially when watching videos in Firefox.

cwegener:
> The resulting CPU usage of damage tracking is VERY high, especially when watching videos in Firefox.

Further testing indicates that the high CPU usage might be isolated to Firefox (or other web browsers).
Using `mpv` for video playback does not exhibit the high CPU usage.

mahkoh:
Seems to be caused by [`d8d6be1` (#223)](https://github.com/mahkoh/jay/pull/223/commits/d8d6be1ef3e79a59ef9cfcd21eaf281a79ccfbf1) which might cause firefox to render at an uncapped frame rate. Maybe this is caused by modifying surface state from multiple threads without synchronization although I don't see an immediate connection.

I've not noticed this in any of the applications that I run normally (chromium, mpv, alacritty, jetbrains IDEs).

cwegener:
Update:

It's ONLY Firefox. Chromium is not affected.

And it looks like it's just because Hardware Video offloading breaks in Firefox after this change (but doesn't break in Chromium).

Might be an Intel driver issue?

I'll try with AMD iGPU and see if it's the same problem

mahkoh:
https://gitlab.freedesktop.org/wayland/wayland/-/issues/481

cwegener:
Just confirming that this also happens on my AMD Ryzen 7 PRO 4750U with Radeon RX Vega 8 iGPU

cwegener:
> And it looks like it's just because Hardware Video offloading breaks in Firefox after this change (but doesn't break in Chromium).

`HARDWARE_VIDEO_DECODING` still shows as 'available / force_enabled' in `about:support` with damage tracking, so it indeed looks to be the issue you identified.

mahkoh:
I'll prepare a PR that fixes the issue with a firefox-specific workaround that restores the previous behavior without precise damage tracking. Let me know if you find any other applications that have this problem.

mahkoh:
Please try https://github.com/mahkoh/jay/pull/224

cwegener:
> Please try #224

I did the same YouTube video playback tests.

This works well on my old Intel i7 test system. (daily driver when at home).

CPU utilization curves sometimes are even 1 or 2% points lower than with 579f93d 

(Now I want to compare power efficiency :wink: )

PR #309, opened 2024-10-28
Implement passive cursor damage to fix cursor refresh rate

pac85:
Jay currently allows to specify a minimum cursor refresh rate. When running with VRR above this refresh rate the cursor completely stops moving which is probably not the intended behavior (tested with vrrtest). 

This patch implements a passive update mode which allows to notify the backend about the cursor position having changed without causing a flip therefore deferring the cursor plane update to the next time the damage call is made to the backend.

### DIFF
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -155,6 +155,7 @@ pub trait HardwareCursorUpdate {
 
 pub trait HardwareCursor: Debug {
     fn damage(&self);
+    fn passive_damage(&self);
 }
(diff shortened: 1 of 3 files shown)


mahkoh:
Please test #311

mahkoh:
This should be fixed by #311.

pac85:
Sorry for the late response (I've been rather busy), I can confirm that #311 solves the issue

PR #310, opened 2024-10-28
wl_surface: don't flip when there is no damage

pac85:
This patch removes a piece of code that will cause a kms flip unnecessarily.

I'm not sure what the original intention of this code was, however what it results in is commits that don't come from mesa WSI to be intepreted as frame submissions from the application.

This introduces flips with spurious timing which break VRR. The exact symptom in many games is that the display goes out of sync when moving the mouse (cursor not visible, not sure why exactly moving the mouse causes the game to commit).

Removing the code fixes this.

### DIFF
--- a/src/ifs/wl_surface.rs
+++ b/src/ifs/wl_surface.rs
@@ -1372,11 +1372,6 @@ impl WlSurface {
                 if has_frame_requests {
                     output.global.connector.damage();
                 }
-            } else if has_frame_requests && output.schedule.vrr_enabled() {
-                // Frame requests must be dispatched at the highest possible frame rate.
-                // Therefore we must trigger a vsync of the output as soon as possible.
-                let rect = output.global.pos.get();
-                self.client.state.damage(rect);
             }
         } else {
             if fifo_barrier_set {


mahkoh:
Can you provide a list of affected games? Unfortunately I believe this code has to stay because removing it would significantly increase the latency of desktop applications. But I might be able to tweak it to handle this case better.

pac85:
Once again sorry for the late response, I've tested a few so here is a list:
* [Ironwail](https://github.com/andrei-drexler/ironwail)
* Xonotic
* Sauerbraten

Probably many more. To reproduce the issue:

* limit the game's framerate below refresh rate
- * For ironwail press `~` then type `host_maxfps <max>`
- * For Xonotic `~` then `cl_maxfps <max>`
- * For Sauerbraten `/` then `maxfps <max>`
* get in game, hold the mouse still; display syncs to the game
* move the mouse -> display jumps to max refresh or other framerates. If enabled you also get tearing

pac85:
BTW can you explain the origin al intention behind the check? It seems to assume that if we get a commit after a frame event has been sent that is a new frame, I think it is at least necessary to check for damage

mahkoh:
Please test https://github.com/mahkoh/jay/pull/312.

>BTW can you explain the origin al intention behind the check?

Some clients behave as follows:

1. They don't do anything for some time.
2. They submit a commit containing only a frame request.
3. They start rendering when they receive the reply to the frame request.

Therefore we must reply to new frame requests as soon as possible. In the case of VRR, this means triggering a vblank as soon as possible.

pac85:
Ah I see
I wonder whether it would be more ideal to send a frame event to those clients without actually triggering scanout (I don't think vblank is instantaneous after a flip).
Anyway thanks for explaining, #312 works so closing this

PR #312, opened 2024-10-30
wl_surface: don't request vblank if there are no new frame requests

mahkoh:


pac85:
this solves the issue, thank you!

PR #335, opened 2024-12-26
tree: focus float nodes when switching workspaces

tadeokondrak:
Right now if a workspace only contains a floating node, switching to that workspace will not change the currently focused window (without focus-follows-mouse doing its thing)

### DIFF
--- a/src/tree/workspace.rs
+++ b/src/tree/workspace.rs
@@ -276,6 +276,14 @@ impl Node for WorkspaceNode {
             fs.tl_into_node().node_do_focus(seat, direction);
         } else if let Some(container) = self.container.get() {
             container.node_do_focus(seat, direction);
+        } else if let Some(float) = self
+            .stacked
+            .rev_iter()
+            .find_map(|node| (*node).clone().node_into_float())
+        {
+            if let Some(child) = float.child.get() {
+                child.node_do_focus(seat, direction);
+            }
         }
     }
 
--- a/src/it/tests.rs
+++ b/src/it/tests.rs
@@ -75,6 +75,7 @@ mod t0040_virtual_keyboard;
 mod t0041_input_method;
 mod t0042_toplevel_select;
 mod t0043_destroy_registry;
+mod t0044_stacked_focus;
 
 pub trait TestCase: Sync {
     fn name(&self) -> &'static str;
@@ -137,5 +138,6 @@ pub fn tests() -> Vec<&'static dyn TestCase> {
         t0041_input_method,
         t0042_toplevel_select,
         t0043_destroy_registry,
+        t0044_stacked_focus,
     }
 }
--- a/src/it/tests/t0044_stacked_focus.rs
+++ b/src/it/tests/t0044_stacked_focus.rs
@@ -1,0 +1,40 @@
+use {
+    crate::{
+        it::{test_error::TestError, testrun::TestRun},
+        tree::ToplevelNodeBase,
+    },
+    std::rc::Rc,
+};
+
+testcase!();
+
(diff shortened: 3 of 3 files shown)


src/tree/workspace.rs
```diff
@@ -274,6 +274,12 @@ impl Node for WorkspaceNode {
     fn node_do_focus(self: Rc<Self>, seat: &Rc<WlSeatGlobal>, direction: Direction) {
         if let Some(fs) = self.fullscreen.get() {
             fs.tl_into_node().node_do_focus(seat, direction);
+        } else if let Some(stacked) = self.stacked.first() {
```
mahkoh: Iirc, this list is sorted from bottom to top. Maybe it would be better to focus the topmost window?

src/tree/workspace.rs
```diff
@@ -274,6 +274,12 @@ impl Node for WorkspaceNode {
...
+            if let Some(float) = Rc::clone(&stacked).node_into_float() {
```
mahkoh: Why not `stacked.clone`?

mahkoh:
This looks good in general.

src/tree/workspace.rs
tadeokondrak: `stacked` is a `NodeRef<Rc<>>` but I changed it to `(*stacked).clone()` which seems very slightly cleaner.

src/tree/workspace.rs
tadeokondrak: Agreed

tadeokondrak:
One thing I'm not sure about is if floating containers should be prioritized over tiling ones. Currently it's the floating containers but that feels motivated by the fact that there doesn't yet seem to be a way to focus them with the keyboard.

mahkoh:
I had only reviewed from my phone so far and hadn't caught that your change prioritizes floating windows over tiled ones. I'd prefer to only focus floating windows if all of the existing logic fails. As you say, the inability to navigate such windows with the keyboard alone exists independently workspace focus behavior.

src/tree/workspace.rs
```diff
@@ -274,6 +274,12 @@ impl Node for WorkspaceNode {
     fn node_do_focus(self: Rc<Self>, seat: &Rc<WlSeatGlobal>, direction: Direction) {
         if let Some(fs) = self.fullscreen.get() {
             fs.tl_into_node().node_do_focus(seat, direction);
+        } else if let Some(stacked) = self.stacked.last() {
+            if let Some(float) = (*stacked).clone().node_into_float() {
+                if let Some(child) = float.child.get() {
+                    child.node_do_focus(seat, direction);
+                }
+            }
```
mahkoh: This should probably loop over all stacked nodes and pick the first one that is a float node. For example, if the topmost node is a popup, then this would not have the desired effect.

tadeokondrak: Done

tadeokondrak:
> I'd prefer to only focus floating windows if all of the existing logic fails

Sounds good, updated

src/tree/workspace.rs
```diff
@@ -276,6 +276,14 @@ impl Node for WorkspaceNode {
             fs.tl_into_node().node_do_focus(seat, direction);
         } else if let Some(container) = self.container.get() {
             container.node_do_focus(seat, direction);
+        } else if let Some(float) = self
+            .stacked
+            .rev_iter()
+            .find_map(|node| (*node).clone().node_into_float())
```
tadeokondrak, line 282: This clone can be avoided but it doesn't allocate so I don't think it makes much of a difference

mahkoh:
Thanks!

Issue #342, opened 2025-01-13
libei server sends ei_device v2 event region_mapping_id to clients binding v1


tadeokondrak:
I plan to send a patch to fix this.

PR #346, opened 2025-01-23
ei: only send region_mapping_id to clients binding v2

tadeokondrak:
Tested with my client that was crashing without this

Closes https://github.com/mahkoh/jay/issues/342


mahkoh:
Thanks!

PR #361, opened 2025-02-05
screencast: re-negotiate format/modifier when size changes

mahkoh:
Some formats/modifiers only work up to a certain size. Using 1x1 would
yield combinations that won't actually work for the screencast.

### DIFF
--- a/deploy-notes.md
+++ b/deploy-notes.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Needs jay-compositor release.
+
 # 1.9.0
 
 - Needs jay-compositor release.
--- a/release-notes.md
+++ b/release-notes.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Various bugfixes.
+
 # 1.9.0 (2025-01-27)
 
 This release replaces xkbcommon by the kbvm crate.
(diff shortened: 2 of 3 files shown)


tadeokondrak:
Capturing full outputs and windows that start out above the correct size works, but renegotiating the format on resize doesn't seem to work for OBS (31.0.1). I don't use any other portal clients so I can't tell if it's a limitation on their side here.

mahkoh:
Please try the code I just pushed and post the logs for both jay and the portal. Renegotiation seems to work fine on my system: 

```

[2025-02-05T18:14:54.734Z INFO  jay::portal::ptl_screencast] Negotiated format argb8888 with modifier 0x20000001876bb03 at size 2560x540
[2025-02-05T18:14:54.736Z INFO  jay::portal::ptl_screencast] Negotiated format argb8888 with modifier 0x200000018737b03 at size 2563x540
```

tadeokondrak:
Hmm it's working perfectly now. Sorry about that, maybe the old portal was running for some reason. If I resize the window repeatedly fast the portal pipewire connection dies but that's a separate issue I assume.

mahkoh:
>If I resize the window repeatedly fast the portal pipewire connection dies but that's a separate issue I assume.

I'm unable to reproduce this.

PR #379, opened 2025-02-22
feat(pointer): implement cursor position hint

### LINKED ISSUE #341: zwp_locked_pointer_v1::set_cursor_position_hint is unimplemented
Currently using this which works but is too hacky to PR:

<details>

<summary>Patch</summary>

```diff
diff --git a/src/ifs/wl_seat/zwp_pointer_constraints_v1.rs b/src/ifs/wl_seat/zwp_pointer_constraints_v1.rs
index 2c1b83b8..3267d5b6 100644
--- a/src/ifs/wl_seat/zwp_pointer_constraints_v1.rs
+++ b/src/ifs/wl_seat/zwp_pointer_constraints_v1.rs
@@ -61,6 +61,7 @@ pub struct SeatConstraint {
     pub one_shot: bool,
     pub status: Cell<SeatConstraintStatus>,
     pub ty: ConstraintType,
+    pub cursor_position_hint: Cell<Option<(Fixed, Fixed)>>,
 }
 
 impl SeatConstraint {
@@ -129,6 +130,17 @@ impl SeatConstraint {
         self.region.set(region);
         Ok(())
     }
+
+    fn warp_to_hint(&self) {
+        if let Some((x, y)) = self.cursor_position_hint.get() {
+            let surface_pos = self.surface.buffer_abs_pos.get();
+            self.seat.motion_event_abs(
+                0,
+                Fixed::from_int(surface_pos.x1()) + x,
+                Fixed::from_int(surface_pos.y1()) + y,
+            );
+        }
+    }
 }
 
 pub trait ConstraintOwner {
@@ -221,6 +233,7 @@ impl ZwpPointerConstraintsV1 {
             one_shot,
             status: Cell::new(SeatConstraintStatus::Inactive),
             ty,
+            cursor_position_hint: Cell::new(None),
         }))
     }
 }
diff --git a/src/ifs/wl_seat/zwp_pointer_constraints_v1/zwp_locked_pointer_v1.rs b/src/ifs/wl_seat/zwp_pointer_constraints_v1/zwp_locked_pointer_v1.rs
index 581df25c..0427e7dc 100644
--- a/src/ifs/wl_seat/zwp_pointer_constraints_v1/zwp_locked_pointer_v1.rs
+++ b/src/ifs/wl_seat/zwp_pointer_constraints_v1/zwp_locked_pointer_v1.rs
@@ -23,6 +23,7 @@ impl ZwpLockedPointerV1RequestHandler for ZwpLockedPointerV1 {
     type Error = ZwpLockedPointerV1Error;
 
     fn destroy(&self, _req: Destroy, _slf: &Rc<Self>) -> Result<(), Self::Error> {
+        self.constraint.warp_to_hint();
         self.constraint.detach();
         self.constraint.client.remove_obj(self)?;
         Ok(())
@@ -30,9 +31,10 @@ impl ZwpLockedPointerV1RequestHandler for ZwpLockedPointerV1 {
 
     fn set_cursor_position_hint(
         &self,
-        _req: SetCursorPositionHint,
+        req: SetCursorPositionHint,
         _slf: &Rc<Self>,
     ) -> Result<(), Self::Error> {
+        self.constraint.cursor_position_hint.set(Some((req.surface_x, req.surface_y)));
         Ok(())
     }
```

</details>

nnyyxxxx:
fixes #341

### DIFF
--- a/src/ifs/wl_seat/zwp_pointer_constraints_v1.rs
+++ b/src/ifs/wl_seat/zwp_pointer_constraints_v1.rs
@@ -67,6 +67,7 @@ pub struct SeatConstraint {
 impl SeatConstraint {
     pub fn deactivate(&self) {
         if self.status.get() == SeatConstraintStatus::Active {
+            let was_locked = self.ty == ConstraintType::Lock;
             self.seat.constraint.take();
             if let Some(owner) = self.owner.get() {
                 owner.send_disabled();
@@ -75,6 +76,13 @@ impl SeatConstraint {
                 self.status.set(SeatConstraintStatus::TerminallyDisabled);
             } else {
                 self.status.set(SeatConstraintStatus::Inactive);
+            }
+            if was_locked {
+                if let Some((hint_x, hint_y)) = self.cursor_hint.get() {
+                    let surface_pos = self.surface.buffer_abs_pos.get();
+                    let (x, y) = (hint_x + surface_pos.x1(), hint_y + surface_pos.y1());
+                    let _ = self.seat.set_pointer_cursor_position(x, y);
+                }
             }
             self.clear_cursor_hint();
         }
--- a/src/ifs/wl_seat/zwp_pointer_constraints_v1.rs
+++ b/src/ifs/wl_seat/zwp_pointer_constraints_v1.rs
@@ -61,6 +61,7 @@ pub struct SeatConstraint {
     pub one_shot: bool,
     pub status: Cell<SeatConstraintStatus>,
     pub ty: ConstraintType,
+    pub cursor_hint: Cell<Option<(Fixed, Fixed)>>,
 }
 
 impl SeatConstraint {
@@ -75,6 +76,7 @@ impl SeatConstraint {
             } else {
                 self.status.set(SeatConstraintStatus::Inactive);
             }
+            self.clear_cursor_hint();
         }
     }
 
@@ -128,6 +130,14 @@ impl SeatConstraint {
         let region = get_region(&self.client, region)?;
         self.region.set(region);
         Ok(())
+    }
+
+    pub fn set_cursor_hint(&self, x: Fixed, y: Fixed) {
+        self.cursor_hint.set(Some((x, y)));
+    }
+
+    pub fn clear_cursor_hint(&self) {
+        self.cursor_hint.set(None);
     }
 }
 
@@ -221,6 +231,7 @@ impl ZwpPointerConstraintsV1 {
             one_shot,
             status: Cell::new(SeatConstraintStatus::Inactive),
             ty,
+            cursor_hint: Cell::new(None),
         }))
     }
 }
--- a/src/ifs/wl_seat/zwp_pointer_constraints_v1.rs
+++ b/src/ifs/wl_seat/zwp_pointer_constraints_v1.rs
@@ -80,8 +80,13 @@ impl SeatConstraint {
             if was_locked {
                 if let Some((hint_x, hint_y)) = self.cursor_hint.get() {
                     let surface_pos = self.surface.buffer_abs_pos.get();
+                    let surface_rect = self.surface.extents.get();
                     let (x, y) = (hint_x + surface_pos.x1(), hint_y + surface_pos.y1());
-                    let _ = self.seat.set_pointer_cursor_position(x, y);
+                    if surface_rect.contains(hint_x.round_down(), hint_y.round_down()) {
+                        if self.seat.cursor_user_group.active().unwrap().get().is_none() {
+                            let _ = self.seat.set_pointer_cursor_position(x, y);
+                        }
+                    }
                 }
             }
             self.clear_cursor_hint();
--- a/src/ifs/wl_seat/zwp_pointer_constraints_v1.rs
+++ b/src/ifs/wl_seat/zwp_pointer_constraints_v1.rs
@@ -82,8 +82,10 @@ impl SeatConstraint {
     pub fn set_cursor_hint(&self, x: Fixed, y: Fixed, apply: bool) {
         self.cursor_hint.set(Some((x, y)));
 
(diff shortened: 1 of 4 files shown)


mahkoh:
I don't think the hint should be considered in the "event-handling" path. For example, in `motion_event_abs`, the user has explicitly told us where to put the cursor. The hint should be irrelevant in that case.

There are currently the following cases in which `SeatConstraint::deactivate` is called and where we might want to look at the hint:

1. Absolute motion events
2. The surface loses pointer focus
3. The lock constraint is destroyed
4. The user explicitly invokes the `disable_pointer_constraint` action
5. When starting a toplevel drag

I already discussed 1) above.

The case of 5) looks quite complicated and the overlap of clients using both locked pointers and toplevel drag is probably zero. Better not to get fancy here.

This leaves 2), 3), and 4):

https://github.com/mahkoh/jay/blob/5625b0585f406f44e2739ce8ad0f55ba368f3566/src/ifs/wl_seat/event_handling.rs#L1270-L1279

https://github.com/mahkoh/jay/blob/5625b0585f406f44e2739ce8ad0f55ba368f3566/src/ifs/wl_seat/zwp_pointer_constraints_v1.rs#L121-L125

https://github.com/mahkoh/jay/blob/5625b0585f406f44e2739ce8ad0f55ba368f3566/src/ifs/wl_seat.rs#L452-L461

In those cases I think it makes sense to apply the hint. But only if the cursor was invisible before then.

src/ifs/wl_seat/zwp_pointer_constraints_v1.rs
```diff
@@ -75,6 +77,14 @@ impl SeatConstraint {
             } else {
                 self.status.set(SeatConstraintStatus::Inactive);
             }
+            if was_locked {
+                if let Some((hint_x, hint_y)) = self.cursor_hint.get() {
+                    let surface_pos = self.surface.buffer_abs_pos.get();
+                    let (x, y) = (hint_x + surface_pos.x1(), hint_y + surface_pos.y1());
+                    let _ = self.seat.set_pointer_cursor_position(x, y);
+                }
```
mahkoh: This code should only run

- in the cases I mentioned in my previous comment,
- if the hint is within the surface, and
- if the cursor is currently invisible.

ghost: how exactly would you check if the cursor is invisible?

ghost: [70fa759](https://github.com/mahkoh/jay/pull/379/commits/70fa75907f963b2e118714a456eebdc58b1f6dd6) ive added a few new checks here according to your requests, we're checking if there is no active cursor via:    if self.seat.cursor_user_group.active().unwrap().get().is_none() but i'm not sure if thats correct

src/ifs/wl_seat/zwp_pointer_constraints_v1.rs
```diff
@@ -75,6 +77,19 @@ impl SeatConstraint {
             } else {
                 self.status.set(SeatConstraintStatus::Inactive);
             }
+            if was_locked {
+                if let Some((hint_x, hint_y)) = self.cursor_hint.get() {
+                    let surface_pos = self.surface.buffer_abs_pos.get();
+                    let surface_rect = self.surface.extents.get();
+                    let (x, y) = (hint_x + surface_pos.x1(), hint_y + surface_pos.y1());
+                    if surface_rect.contains(hint_x.round_down(), hint_y.round_down()) {
+                        if self.seat.cursor_user_group.active().unwrap().get().is_none() {
```
mahkoh: I think the correct way to check this is

```rust
impl CursorUser {
    pub fn is_visible(&self) -> bool {
        self.is_active() && self.cursor.is_some()
    }

    pub fn is_invisible(&self) -> bool {
        !self.is_visible()
    }
}
```

and then

```rust
self.seat.pointer_cursor.is_invisible()
```

here.

src/ifs/wl_seat/zwp_pointer_constraints_v1.rs
```diff
@@ -61,11 +61,13 @@ pub struct SeatConstraint {
     pub one_shot: bool,
     pub status: Cell<SeatConstraintStatus>,
     pub ty: ConstraintType,
+    pub cursor_hint: Cell<Option<(Fixed, Fixed)>>,
 }
 
 impl SeatConstraint {
     pub fn deactivate(&self) {
```
mahkoh, line 93: This function needs to take a parameter that says whether the hint should be applied.

src/ifs/wl_seat/zwp_pointer_constraints_v1.rs
ghost: done.

src/ifs/wl_seat/zwp_pointer_constraints_v1.rs
ghost: done.

mahkoh: It's not done.

ghost: > It's not done.

oh srry... i've been too caught up with a lot of things, minor details abliss

ghost: uhm. now that i look again, this is done right? its only being applied when:

apply is true
the constraint is active
the constraint has type lock

or am i missing something here?

mahkoh: Please consider how your code implements the requirements from https://github.com/mahkoh/jay/pull/379#issuecomment-2677703989.

ghost: uh.. looks good to me, though im not that good at digesting large bodies of text

PR #406, opened 2025-03-14
ei: Fix flipped scroll axes

tadeokondrak:
This allows for emulating ordinary vertical scrolling with libei. I think horizontal scrolling might be broken due to [this](https://github.com/mahkoh/jay/blob/efd9e38c28a0f902947858277b3b026fd3d2e44f/src/ifs/jay_seat_events.rs#L98) but I haven't looked into that yet.

### DIFF
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -280,7 +280,7 @@ pub enum KeyState {
     Pressed,
 }
 
-#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Linearize)]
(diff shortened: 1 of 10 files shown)


mahkoh:
I think the better solution is to remove the uses of the magic numbers 0 and 1 and replace them with the VERTICAL_SCROLL and HORIZONTAL_SCROLL constants. Outgoing EI messages are broken in the same way as far as I can tell.

tadeokondrak:
I'm running the new code now and it still solves the original issue. Works for me

mahkoh:
Thanks!

Issue #324, opened 2024-11-23
Portal does not work with dbus-run-session


tadeokondrak:
dbus-run-session is some sort of namespacing for a dbus session bus - I'm not exactly sure, I just cargo-cult copied the habit of using `dbus-run-session sway` and copied that over to when I switched to jay.

Jay's dbus library uses `$XDG_RUNTIME_DIR/bus` instead of parsing `$DBUS_SESSION_BUS_ADDRESS` - this was confusing to figure out. I actually don't need this to be resolved in Jay now that I stopped using dbus-run-session, but it's a possible improvement.

mahkoh:
Thanks for reporting this.

If someone wants to submit a patch to fix this issue, I'd be happy to review it. I believe the [zbus crate](https://github.com/dbus2/zbus) can serve as a reference.

PR #417, opened 2025-04-04
dbus: Parse DBUS_SESSION_BUS_ADDRESS instead of using XDG_RUNTIME_DIR

tadeokondrak:
The old way of using XDG_RUNTIME_DIR doesn't work for multiple sessions using dbus-run-session.

Fixes https://github.com/mahkoh/jay/issues/324

### DIFF
--- a/src/dbus.rs
+++ b/src/dbus.rs
@@ -139,6 +139,8 @@ pub enum DbusError {
 }
 efrom!(DbusError, IoUringError);
 
+const DBUS_SESSION_BUS_ADDRESS: &str = "DBUS_SESSION_BUS_ADDRESS";
+
 pub struct Dbus {
     eng: Rc<AsyncEngine>,
     ring: Rc<IoUring>,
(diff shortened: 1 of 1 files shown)


src/dbus.rs
```diff
@@ -149,12 +150,63 @@ pub struct Dbus {
...
+                    | b'/'
+                    | b'.'
+                    | b'\\'
+                    | b'*' => {
+                        unescaped.push(c);
+                    }
+                    b'%' => {
+                        let hi = (bytes.next()? as char).to_digit(16)?;
+                        let lo = (bytes.next()? as char).to_digit(16)?;
+                        unescaped.push((hi << 4 | lo) as u8);
```
tadeokondrak, line 174: ~~Realizing I haven't tested escaped characters at all yet. Will resolve this comment once I do that~~ Done

src/dbus.rs
```diff
@@ -149,12 +150,63 @@ pub struct Dbus {
...
+                    }
+                    _ => return None,
+                }
+            }
+            String::from_utf8(unescaped).ok()
+        }
+        let user_path = 'path: {
+            let Some(addr) = std::env::var(DBUS_SESSION_BUS_ADDRESS).ok() else {
+                log::warn!("{DBUS_SESSION_BUS_ADDRESS} is not set");
+                break 'path None;
```
mahkoh, line 187: Please add a fallback to the original logic here.

mahkoh:
Looks good to me. I'll test it tomorrow.

src/dbus.rs
tadeokondrak: I can do that. ~~It doesn't seem necessary, though - zbus doesn't do it, so maybe it's only necessary for broken setups.~~

tadeokondrak: Sorry, I was wrong in my last comment. Will definitely add the fallback

mahkoh:
Thanks!

PR #419, opened 2025-04-05
wl_drm: hide global if invalid modifier is not supported

### LINKED ISSUE #418: Xwayland/OpenGL: Unable to create prime buffer on non-primary renderer (DRI-PRIME)
Hi,

When trying to run an OpenGL application on XWayland on the non-primary renderer, the Xwayland server crashes with:

```text
X connection to :500 broken (explicit kill or server shutdown).
X connection to :500 broken (explicit kill or server shutdown).
```

And the Jay log reports:

```text
Xwayland: XWAYLAND: wl_display#1: error 3: An error occurred while processing a request: Could not process a `wl_drm#13.create_prime_buffer` request: Could not import the buffer: The modifier is not supported
```

**Test Case**

```
DRI_PRIME=1 glxgears
DRI_PRIME=1 eglgears_x11
```

**Environment/Config**

```
Primary-renderer: iGPU (AMD Radeon 680M)
Secondary-renderer: dGPU (AMD Radeon RX 7600S)
GraphicsApi: Vulkan
ExcplicitSync: true (tested without)
DirectScanout: primary & secondary (tested without)
Jay Version: Master branch (commit: c6ca7c228c5ce0b29ffb19dc4726793d3b1c9497)
```

**Workarounds**

I tested the following workarounds:

1. Disable direct-scan out -> No impact
2. Disable explicit sync -> No impact
3. Set output format on both cards to be the same-> No impact (Note: I only tested one of them!)
4. Set the graphics API on the primary renderer to OpenGL -> Works!
5. Set the dGPU to primary renderer -> Works, but same issue now rendering on the iGPU 

For sanity check, I also tested the same kind of setup on Sway using `WLR_RENDERER=vulkan` and it works there

Let me know if there are other workarounds/tests I should checkout in terms of debugging efforts.

Thanks in advance!

mahkoh:
Fixes #418

### DIFF
--- a/src/gfx_api.rs
+++ b/src/gfx_api.rs
@@ -795,6 +795,10 @@ pub trait GfxContext: Debug {
     fn supports_color_management(&self) -> bool {
         false
     }
+
+    fn supports_invalid_modifier(&self) -> bool {
+        false
+    }
 }
 
 #[derive(Clone, Debug)]
--- a/src/gfx_apis/gl/renderer/context.rs
+++ b/src/gfx_apis/gl/renderer/context.rs
@@ -347,4 +347,8 @@ impl GfxContext for GlRenderContext {
     ) -> Result<Rc<dyn GfxBlendBuffer>, GfxError> {
         Err(GfxError(Box::new(RenderError::NoBlendBuffer)))
     }
+
+    fn supports_invalid_modifier(&self) -> bool {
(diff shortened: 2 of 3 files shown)


tskaar:
Confirmed working on my end, thanks for the quick fix! :rocket:

Issue #407, opened 2025-03-16
No official community for Jay


disluckyguy:
Hi! I found out about this project and I'm really interested and amazed by it!

But I found that there's no official community (like a discord server, matrix channel, subreddit, ...etc)
Having a place like this will really help people who are interested (like me) to be able to more easily contact the developer(s), find others who like Jay, and in general is much easier to talk in than through issues.

Are there any plans to make one?

Thank you and sorry if this is not suitable to write as an issue.

mahkoh:
I am at the moment not interested in managing such a place. If people have questions, they can ask them in this issue tracker.

However, if people want to create an unofficial discord server, I would not be opposed to that and would add a link to it to the readme.

disluckyguy:
Thank you for answering,

> I am at the moment not interested in managing such a place
Would you be fine joining and communicating there, without moderating?

disluckyguy:
I have made a discord server as a start, if anyone wants to join:

https://discord.gg/SbR5UFk2

However I do not suggest you put it on readme until I set everything up

khyperia:
The discord link here, as well as the one in #413, result in this for me. Any chance I/we could get a valid link?

![Image](https://github.com/user-attachments/assets/63f0744f-bfa2-467f-9d46-f1ef5e83c484)

disluckyguy:
Hi @khyperia, here's a working one
https://discord.gg/MxkYChd8,
Besides, is there any way to have a permanent link?

PR #424, opened 2025-04-19
Update README.md discord link to an unexpirable one

disluckyguy:
The old link expired after a week from creation, I've set the links to never expire, this one is the new one
In response to a complaint in #407


mahkoh:
Thanks!

PR #446, opened 2025-04-26
jay-config: add Seat.get_keyboard_output and Connector.workspaces

khyperia:
This is a bit of a combined PR with a bunch of separate changes. Please let me know if you'd like me to split each change into individual PRs.

I've been working on my own config setup and found that two pieces of functionality were missing to be able to create my config:

- Get which workspaces are on which outputs, and which workspace is actively being displayed on each output.
- Get which workspace the active keyboard focus is on (not the mouse position)

There's two different ways of doing the first point that I was debating between (they expose equivalent functionality): either `Workspace.get_output` that returns a Connector and a bool for whether it's the active workspace, or `Connector.workspaces` that returns the active workspace, and the list of workspaces on that output. I opted for the latter, mostly arbitrarily.

Additionally, there's some more changes I did in this PR:

- `handle_get_seat_workspace` (called by `Seat.get_workspace`) would fail if `workspaces_by_name` has not been initialized yet. I extracted the common initialization code of `handle_get_workspaces` and `handle_get_workspace` into `get_workspace_by_name`, and am using that same function in the new `handle_get_seat_keyboard_workspace` and `handle_get_connector_workspaces`.

Finally, I'm not 100% sure on the implementation of `WlSeatGlobal.get_keyboard_output` - I don't know if `node_toplevel` is the right function to call there.

### DIFF
--- a/jay-config/src/video.rs
+++ b/jay-config/src/video.rs
@@ -3,7 +3,7 @@
 use {
     crate::{
         _private::WireMode,
-        PciId,
+        PciId, Workspace,
         video::connector_type::{
             CON_9PIN_DIN, CON_COMPONENT, CON_COMPOSITE, CON_DISPLAY_PORT, CON_DPI, CON_DSI,
             CON_DVIA, CON_DVID, CON_DVII, CON_EDP, CON_EMBEDDED_WINDOW, CON_HDMIA, CON_HDMIB,
@@ -300,6 +300,21 @@ impl Connector {
     /// This has no effect unless the vulkan renderer is used.
     pub fn set_brightness(self, brightness: Option<f64>) {
         get!().connector_set_brightness(self, brightness);
+    }
+
+    /// Get the currently visible/active workspace.
+    ///
+    /// If this connector is not connected, or is there no active workspace, returns a
+    /// workspace whose `exists()` returns false.
+    pub fn active_workspace(self) -> Workspace {
+        get!(Workspace(0)).get_connector_active_workspace(self)
+    }
+
+    /// Get all workspaces on this connector.
+    ///
+    /// If this connector is not connected, returns an empty list.
+    pub fn workspaces(self) -> Vec<Workspace> {
+        get!().get_connector_workspaces(self)
     }
 }
 
--- a/jay-config/src/_private/client.rs
+++ b/jay-config/src/_private/client.rs
@@ -417,6 +417,12 @@ impl Client {
         workspace
     }
 
+    pub fn get_seat_keyboard_workspace(&self, seat: Seat) -> Workspace {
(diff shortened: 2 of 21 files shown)


khyperia:
For context, the functionality that I want to implement in my personal config is:

- A keybind that focuses "the next monitor" - I only have 2, so in my case, it's "focus the other one"
  - same for "move window to the next monitor"
- A keybind that focuses the previous/next workspace in the current display's workspace list: if the workspaces [1, 3, 4] are on my first monitor, with [1] being displayed, and I press the keybind, [3] should be focused, then [4], then back to [1].
  - same for "move window to previous/next workspace"

Adding `Seat.get_keyboard_output` and `Connector.workspaces` is enough to implement those.

(I'd be happy to add these to `toml-config` as well, but I figured these commands are a bit niche and them just living in my `config.so` is good enough)

mahkoh:
>Finally, I'm not 100% sure on the implementation of WlSeatGlobal.get_keyboard_output - I don't know if node_toplevel is the right function to call there.

I don't think that would work but I've fixed it.

mahkoh:
LGTM

khyperia:
I've pushed an additional commit: when closing a window, it now focuses the OutputNode of `seat.get_output`, rather than the root DisplayNode. This means that `Seat.get_keyboard_workspace` continues to function after closing the last window on a Workspace, instead of returning None/invalid.

I have design opinions around how exactly focus and how these types of actions work, though - I'll open a new issue for discussion instead of clogging this PR (which I hope can now land as-is! Let me know if there's any additional things I can do to help this land.)

mahkoh:
I don't know if this is correct. Output nodes never have the keyboard focus and there is no code to ensure that the focus is released when the output is removed. However, if the code in the commit is actually what you want, then you can just Seat::get_workspace if Seat::get_keyboard_workspace does not exist.

khyperia:
right! I have reverted that commit, and will look into it / maybe file a follow-up PR - but not cluttering this one with that change, so that this PR can get in.

jay-config/src/video.rs
```diff
@@ -301,6 +301,14 @@ impl Connector {
     pub fn set_brightness(self, brightness: Option<f64>) {
         get!().connector_set_brightness(self, brightness);
     }
+
+    /// Get both the currently visible workspace and all workspaces on this connector.
+    ///
+    /// If this connector is not connected, or is there no active workspace, returns a
+    /// workspace whose `exists()` returns false.
+    pub fn workspaces(self) -> (Workspace, Vec<Workspace>) {
```
mahkoh: I think this should be split into two functions, one that returns the visible workspace and one that returns all workspaces. Same for the IPC messages.

mahkoh:
Thanks!

PR #449, opened 2025-04-28
implement new setting: relative behavior mode

### LINKED ISSUE #448: Add option for more keyboard-centric behavior rather than mouse-centric
I'm going to implement this myself (unless others would like to), I'd just like to discuss some design decisions:

Right now, Jay has a fairly mouse-centric design. If you have two monitors, with a window focused on one monitor, and your mouse located on another:

- Opening a new window puts it on the mouse monitor, not the focused-window monitor
- Closing a window on the focused-window monitor (with keyboard input) refocuses onto a window on the mouse monitor, not the monitor the closed window used to be on
- Creating a new workspace adds the workspace to the mouse monitor, not the focused-window monitor
- Perhaps a couple other places I'm missing

I prefer having behavior centered around the focused window, rather than location of the mouse, so:

If this is intentional design, would a PR be accepted with me adding an option to make these things follow a "focused window"-centric design rather than mouse-centric?

Or if it's unintentional, I implement no option at all, and all these things just follow the focused window rather than mouse?

khyperia:
Partially fixes #448. I have no idea what a good name for this is, I just guessed "relative behavior mode".

This PR gets most of the way there, but has a glaring hole: as discussed in #446, Jay doesn't really have support for setting keyboard_node to an OutputNode or a WorkspaceNode. This means that for workspaces with no windows, the focus gets "lost": for example, `seat.show_workspace` does nothing when passed a workspace with no windows, which breaks my expected functionality - if I call `seat.show_workspace`, then open a window, I expect the window to be opened on that workspace.

So, I intend to implement that missing functionality in a follow-up PR, as I'm guessing you'd like such a tangly PR (implementing defocusing selected OutputNodes when the output is disconnected, etc.) to be separated from this option/feature PR. Let me know if you'd instead like that to be in this PR as well, though!

(Thanks for taking the time to review my PRs and look at my issues! I really appreciate it, I'm really enjoying using Jay, it's great)


jay-config/src/input.rs
```diff
@@ -490,6 +495,15 @@ pub enum FocusFollowsMouseMode {
...
 }
 
+/// Defines where relative actions are performed by default.
+#[derive(Serialize, Deserialize, Copy, Clone, Debug, Hash, Eq, PartialEq)]
+pub enum RelativeBehaviorMode {
+    /// New windows, workspaces, etc. will be created on the display the cursor is on.
+    Cursor,
+    /// New windows, workspaces, etc. will be created on the display the focus is on (which window is highlighted).
+    Focus,
+}
```
mahkoh: I find this a bit hard to understand. I'd prefer to have specialized settings that do one thing only. For example

```rust
pub enum WindowOpenLocation {
    Cursor,
    Keyboard,
}

/// Where to focus when closing a window.
pub enum FocusFallback {
    CursorOutput,
    KeyboardOutput,
}
```

mahkoh:
>This PR gets most of the way there, but has a glaring hole: as discussed in https://github.com/mahkoh/jay/pull/446, Jay doesn't really have support for setting keyboard_node to an OutputNode or a WorkspaceNode.

I think the solution is to not use the keyboard node at all in `get_keyboard_output`. Instead, whenever `WlSeatGlobal::keyboard_node` is updated, you at that point look up the output of that node and store the `OutputGlobalOpt` of the output in a new field `last_keyboard_output`. If the node does not have an output (only the root node), then that field is not updated.

jay-config/src/input.rs
khyperia: Hmm, I don't agree:

The setting here is for every place the codebase calls `WlSeatGlobal.get_output`. That would mean a setting for:

- Where the new workspace is put when set_workspace is called on a new workspace
- Where the new workspace is put when show_workspace is called on a new workspace
- Which workspace is used for `Seat.move_to_output`
- Which window is focused when closing a window
- Which workspace a new window is opened on
- Which workspace new floating windows are opened on
- I honestly don't 100% understand `ZwlrLayerShellV1.get_layer_surface`'s use of get_output, and would need a bit of help coming up with a name/description for a setting for that.

Granted, a couple of those could be merged, but that's still a lot of separate settings. In my opinion it's highly unlikely that a user will want to configure different behavior for each one individually, and will set them all to the same value. Additionally, any new uses of `WlSeatGlobal.get_output` will need to create a new setting, and then users will have to add that new setting to their configuration file for every update that includes a new use of get_output.

It seems better to have one global setting for whether Jay tends to follow mouse-centric behavior or keyboard-centric behavior.

mahkoh: Makes sense.

I'd still want this setting to be renamed to something like `FallbackOutputMode` ~~and be a setting on a `Seat`~~ it already is.

khyperia: Lovely, thanks! And yes, "relative behavior mode" is a terrible name I just guessed at, `FallbackOutputMode` is much better, thanks, haha.

khyperia:
> look up the output of that node and store the `OutputGlobalOpt` of the output in a new field `last_keyboard_output`

I ran into some complications while implementing this:

- Nodes can move outputs. So, whenever a node moves outputs, a check needs to be added to update `last_keyboard_output` if the moved window is the current `keyboard_node`. Maybe that's just `tl_workspace_output_changed`? Unsure if non-toplevels also need handling.
- When calling `Seat.show_workspace` on an empty workspace (conceptually "focusing" the workspace), then creating a new window, I want the window on that workspace. That would mean setting `last_keyboard_output` to that workspace's output, without `keyboard_node` ever actually pointing to that output, which feels a little dirty with the naming `last_keyboard_output`.
  - ... and also things like moving empty workspaces between outputs technically should be detected and move `last_keyboard_output` as well (if that workspace is "focused"), but that's enough of an edge case that I feel like that's fine to miss.

I'm totally fine with doing those, I just wanted to run them past you to see how you feel about those vs. adding support for having a WorkspaceNode and/or an OutputNode be the keyboard_node (either would work fine).

(Looking through the code, it looks like adding support for focusing WorkspaceNodes would be easiest, as they're only destroyed in one place in code right now AFAIK - `OutputNode.show_workspace`, which already had code handling references to the destroyed WorkspaceNode stored elsewhere. But again, up to you, totally reasonable if you'd rather do `last_keyboard_output`!)

mahkoh:
I was busy with #451 but I'll look at this next week.

khyperia:
Thanks! (And to be clear, I haven't implemented either of `last_keyboard_output` or focusing `WorkspaceNode`s yet, that push was just renaming the setting - half-waiting for you to give feedback on which is better, but also half-waiting for me to have time/energy to fully 100% implement either approach)

mahkoh:
How about this?

khyperia:
> How about this?

Mmm! Setting it to the _old_ keyboard_node's output (not the new keyboard_node being set) is very subtle and took me a while to see, that's pretty clever.

But yeah, I'm not sure how to handle focusing an empty workspace with this way of handling it... unfortunately I'm extremely fried from coding at my day job right now :P

mahkoh:
I'm also fine with merging this as is without solving the empty workspace issue.

khyperia:
Update: The one place workspaces are destroyed (`OutputNode::show_workspace`) already handles swapping focus if a workspace node is selected directly, by nature of how the code is written (`collect_kb_foci2` picks up on the fact that the old WorkspaceNode is focused, and then the seat's focus gets refocused via `ws.node_do_focus(Direction::Unspecified)` on the newly shown workspace).

So, implementing focusing empty workspaces is a matter of two lines of code. I've added that as a commit in this PR, feel free to force-push-remove it and we can continue in another PR.

jay-config/src/input.rs
```diff
@@ -650,6 +655,18 @@ pub enum FocusFollowsMouseMode {
     False,
 }
 
+/// Defines which output is used when no particular output is specified.
+///
+/// This configures where to place a newly opened window or workspace, what window to focus when a
+/// window is closed, which workspace is moved with [`Seat::move_to_output`], and similar actions.
+#[derive(Serialize, Deserialize, Copy, Clone, Debug, Hash, Eq, PartialEq)]
+pub enum FallbackOutputMode {
```
mahkoh, line 666:
```suggestion
#[derive(Serialize, Deserialize, Copy, Clone, Debug, Hash, Eq, PartialEq)]
#[non_exhaustive]
pub enum FallbackOutputMode {
```

src/tree/output.rs
```diff
@@ -1495,6 +1496,8 @@ impl OutputNode {
             if c.node_visible() {
                 c.node_do_focus(seat, direction);
             }
+        } else if seat.fallback_output_mode() == FallbackOutputMode::Focus {
+            seat.focus_node(ws);
```
mahkoh: I don't like this. I would like this to just be

```suggestion
        } else {
            seat.focus_node(ws);
```

All of this logic was added after you opened this MR. The focus function can now move the focus across outputs.

Currently this has the limitation that you cannot move the focus to an output that only contains floating windows. So let's say that your focus is on the left output and the right output contains only a floating window, then `focus-right` will not move the focus to that output.

However, with the version that I've suggested above, it would focus the workspace and then you could use `focus-above` to move the focus to the floating window.

----------

The reason I've added the `if` is that I don't like that there is no indication that a workspace is focused. If you have two empty workspaces and one has the focus, there is no way to know which one has the focus.

I think the easiest solution is to change the background color of the workspace in the bar if it has the focus. But the theming code is a mess. The workspace background color of the visible workspace is currently governed by these two settings:

```rust
        /// The title background color of a focused window.
        ///
        /// Default: `#285577`.
        const 02 => FOCUSED_TITLE_BACKGROUND_COLOR,
        /// The title background color of a focused window that might be captured.
        ///
        /// Default: `#772831`.
        const 13 => CAPTURED_FOCUSED_TITLE_BACKGROUND_COLOR,
```

We would have to add two new settings `FOCUSED_WORKSPACE_TITLE_BACKGROUND_COLOR` and `CAPTURED_FOCUSED_WORKSPACE_TITLE_BACKGROUND_COLOR` which would govern the workspace title background _if_ the workspace itself is the keyboard node.

mahkoh: Or there could be some other indicator. Like changing the color of the bar or the bar separator.

khyperia: Yeah... on a related note, ext-workspace-v1 doesn't have a distinction between "visible" and "focused". Waybar's ext-workspace-v1 backend [checks for the Active state](https://github.com/Alexays/Waybar/blob/06484547d182057159c4bd58d14871c0b7a7f3dc/src/modules/ext/workspace_manager.cpp#L394-L397), but its Sway IPC workspace backend [checks for the distinction between "visible" and "focused"](https://github.com/Alexays/Waybar/blob/06484547d182057159c4bd58d14871c0b7a7f3dc/src/modules/sway/workspaces.cpp#L298-L307), which styles the focused workspace icon differently. Perhaps I'll look into proposing a change to ext-workspace-v1 to add that.

src/tree/workspace.rs
```diff
@@ -374,6 +374,8 @@ impl Node for WorkspaceNode {
             if let Some(child) = float.child.get() {
```
khyperia: Can `float.child` be `None`? I could refactor this to properly fall back to focusing the workspace if it's `None`, but from a few moments of testing I can't figure out a way for it to be None.

src/tree/output.rs
khyperia: Would you like me to work on focused workspace theming in this PR or look into it in a future PR?

(I agree with you that while it'd be nice to not have this check for the fallback_output_mode==Focus, the check should be present until there's a better indication of focused workspace. FWIW, I've been running this patch as my daily driver and I'm happy with the current state of it, especially with your tweaks.)

mahkoh: >Perhaps I'll look into proposing a change to ext-workspace-v1 to add that.

Sounds like a good idea.

mahkoh: I've made it so that the separator gets the focused-title color when the workspace is focused. And I've removed the check. This should be fine.

src/tree/workspace.rs
mahkoh: I don't think it can be None. But you can replace this by `&& let Some(child) = ...` if you want.

mahkoh:
I've pushed another version with the workspace-focus highlighting. I don't quite remember if I had any other concerns. I'll have to do another review. But this looks fine conceptually.

Please rebase your branch to integrate my `some ideas` commit into yours.

khyperia:
> Please test the current version.

Works great!

Thanks for implementing the toml-config stuff as well, you're going above and beyond supporting my silly little desired behavior for me (and hopefully others).

mahkoh:
Thanks!

PR #452, opened 2025-05-05
implement zwlr_foreign_toplevel_management protocol

disluckyguy:
This protocol will provide a way to interact with the compositor windows through wayland, most features are implemented except minimize and maximize, because they're not supported by jay; and set_rectangle.

### DIFF
--- a/src/client/objects.rs
+++ b/src/client/objects.rs
@@ -23,16 +23,15 @@ use {
             wl_output::WlOutput,
             wl_region::WlRegion,
             wl_registry::WlRegistry,
-            wl_seat::{WlSeat, tablet::zwp_tablet_tool_v2::ZwpTabletToolV2, wl_pointer::WlPointer},
+            wl_seat::{tablet::zwp_tablet_tool_v2::ZwpTabletToolV2, wl_pointer::WlPointer, WlSeat},
             wl_surface::{
-                WlSurface,
-                xdg_surface::{XdgSurface, xdg_popup::XdgPopup, xdg_toplevel::XdgToplevel},
+                xdg_surface::{xdg_popup::XdgPopup, xdg_toplevel::XdgToplevel, XdgSurface}, WlSurface
             },
             workspace_manager::ext_workspace_group_handle_v1::ExtWorkspaceGroupHandleV1,
             wp_drm_lease_connector_v1::WpDrmLeaseConnectorV1,
             wp_linux_drm_syncobj_timeline_v1::WpLinuxDrmSyncobjTimelineV1,
             xdg_positioner::XdgPositioner,
-            xdg_wm_base::XdgWmBase,
+            xdg_wm_base::XdgWmBase, zwlr_foreign_toplevel_handle_v1::ZwlrForeignToplevelHandleV1,
         },
         object::{Object, ObjectId},
         utils::{
@@ -40,13 +39,7 @@ use {
             copyhashmap::{CopyHashMap, Locked},
         },
         wire::{
-            ExtDataControlSourceV1Id, ExtForeignToplevelHandleV1Id, ExtImageCaptureSourceV1Id,
-            ExtImageCopyCaptureSessionV1Id, ExtWorkspaceGroupHandleV1Id, JayOutputId,
-            JayScreencastId, JayToplevelId, JayWorkspaceId, WlBufferId, WlDataSourceId, WlOutputId,
-            WlPointerId, WlRegionId, WlRegistryId, WlSeatId, WlSurfaceId, WpDrmLeaseConnectorV1Id,
-            WpImageDescriptionV1Id, WpLinuxDrmSyncobjTimelineV1Id, XdgPopupId, XdgPositionerId,
-            XdgSurfaceId, XdgToplevelId, XdgWmBaseId, ZwlrDataControlSourceV1Id,
-            ZwpPrimarySelectionSourceV1Id, ZwpTabletToolV2Id,
+            ExtDataControlSourceV1Id, ExtForeignToplevelHandleV1Id, ExtImageCaptureSourceV1Id, ExtImageCopyCaptureSessionV1Id, ExtWorkspaceGroupHandleV1Id, JayOutputId, JayScreencastId, JayToplevelId, JayWorkspaceId, WlBufferId, WlDataSourceId, WlOutputId, WlPointerId, WlRegionId, WlRegistryId, WlSeatId, WlSurfaceId, WpDrmLeaseConnectorV1Id, WpImageDescriptionV1Id, WpLinuxDrmSyncobjTimelineV1Id, XdgPopupId, XdgPositionerId, XdgSurfaceId, XdgToplevelId, XdgWmBaseId, ZwlrDataControlSourceV1Id, ZwlrForeignToplevelHandleV1Id, ZwpPrimarySelectionSourceV1Id, ZwpTabletToolV2Id
         },
     },
     std::{cell::RefCell, rc::Rc},
@@ -81,6 +74,8 @@ pub struct Objects {
     pub image_capture_sources: CopyHashMap<ExtImageCaptureSourceV1Id, Rc<ExtImageCaptureSourceV1>>,
     pub foreign_toplevel_handles:
         CopyHashMap<ExtForeignToplevelHandleV1Id, Rc<ExtForeignToplevelHandleV1>>,
+    pub zwlr_foreign_toplevel_handles:
+        CopyHashMap<ZwlrForeignToplevelHandleV1Id, Rc<ZwlrForeignToplevelHandleV1>>,
     pub ext_copy_sessions:
         CopyHashMap<ExtImageCopyCaptureSessionV1Id, Rc<ExtImageCopyCaptureSessionV1>>,
     pub ext_data_sources: CopyHashMap<ExtDataControlSourceV1Id, Rc<ExtDataControlSourceV1>>,
@@ -122,6 +117,7 @@ impl Objects {
             xdg_popups: Default::default(),
             image_capture_sources: Default::default(),
             foreign_toplevel_handles: Default::default(),
+            zwlr_foreign_toplevel_handles: Default::default(),
             ext_copy_sessions: Default::default(),
             ext_data_sources: Default::default(),
             ext_workspace_groups: Default::default(),
@@ -165,6 +161,7 @@ impl Objects {
         self.xdg_popups.clear();
         self.image_capture_sources.clear();
         self.foreign_toplevel_handles.clear();
+        self.zwlr_foreign_toplevel_handles.clear();
         self.ext_copy_sessions.clear();
         self.ext_data_sources.clear();
         self.ext_workspace_groups.clear();
--- a/src/ifs/zwlr_foreign_toplevel_handle_v1.rs
+++ b/src/ifs/zwlr_foreign_toplevel_handle_v1.rs
@@ -1,0 +1,217 @@
+use {
+    super::wl_output::WlOutput,
+    crate::{
+        client::{Client, ClientError},
+        leaks::Tracker,
+        object::{Object, Version},
+        tree::{OutputNode, ToplevelOpt},
+        wire::{ZwlrForeignToplevelHandleV1Id, zwlr_foreign_toplevel_handle_v1::*},
+    },
+    std::rc::Rc,
+    thiserror::Error,
+};
+
+pub struct ZwlrForeignToplevelHandleV1 {
+    pub id: ZwlrForeignToplevelHandleV1Id,
+    pub client: Rc<Client>,
+    pub tracker: Tracker<Self>,
+    pub toplevel: ToplevelOpt,
+    pub version: Version,
+}
+
+impl ZwlrForeignToplevelHandleV1 {
+    fn detach(&self) {
+        if let Some(tl) = self.toplevel.get() {
+            tl.tl_data().manager_handles.remove(&(self.client.id, self.id));
+        }
+    }
+}
+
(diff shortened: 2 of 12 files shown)


mahkoh:
Needs to be formatted with `cargo +nightly fmt`.

src/client/objects.rs
```diff
@@ -81,6 +82,8 @@ pub struct Objects {
     pub image_capture_sources: CopyHashMap<ExtImageCaptureSourceV1Id, Rc<ExtImageCaptureSourceV1>>,
     pub foreign_toplevel_handles:
         CopyHashMap<ExtForeignToplevelHandleV1Id, Rc<ExtForeignToplevelHandleV1>>,
+    pub zwlr_foreign_toplevel_handles:
+        CopyHashMap<ZwlrForeignToplevelHandleV1Id, Rc<ZwlrForeignToplevelHandleV1>>,
```
mahkoh: This field is unused.

src/ifs/zwlr_foreign_toplevel_handle_v1.rs
```diff
@@ -0,0 +1,225 @@
...
+                let ws_outputs = ws_bindigs.get(&self.client.id).map(|hm| hm.values());
+
+                ws_outputs
+                    .map(|mut os| os.any(|o| o.id == req.output))
+                    .unwrap_or(false)
+            });
+
+            if let Some(ws) = ws {
+                toplevel.tl_set_workspace(ws);
+                toplevel.tl_set_fullscreen(true);
```
mahkoh: This is not correct. Instead you should change `tl_set_fullscreen` to take an optional output.

src/ifs/zwlr_foreign_toplevel_handle_v1.rs
```diff
@@ -0,0 +1,225 @@
...
+                            self.send_output_enter(output.clone());
+                        }
+                    }
+                }
+            } else {
+                for output in new_outputs {
+                    self.send_output_enter(output.clone());
+                }
+            }
+        }
```
mahkoh: This is way too complicated. All of these collect calls have to go.

```
if old_output:
    if old_output == new_output:
        return;
    for each old binding:
        send_leave
for each new binding:
    send_enter
```

src/ifs/zwlr_foreign_toplevel_handle_v1.rs
```diff
@@ -0,0 +1,225 @@
...
+        self.client.event(AppId {
+            self_id: self.id,
+            app_id,
+        });
+    }
+
+    pub fn send_state(&self, maximized: bool, minimized: bool, activated: bool, fullscreen: bool) {
+        let mut state: Vec<u32> = vec![];
+        if maximized {
+            state.push(0);
```
mahkoh: All of these numbers need constants.

src/ifs/zwlr_foreign_toplevel_handle_v1.rs
```diff
@@ -0,0 +1,225 @@
...
+
+    pub fn send_app_id(&self, app_id: &str) {
+        self.client.event(AppId {
+            self_id: self.id,
+            app_id,
+        });
+    }
+
+    pub fn send_state(&self, maximized: bool, minimized: bool, activated: bool, fullscreen: bool) {
+        let mut state: Vec<u32> = vec![];
```
mahkoh: This should be an ArrayVec.

src/ifs/zwlr_foreign_toplevel_handle_v1.rs
```diff
@@ -0,0 +1,225 @@
...
+    fn break_loops(&self) {
+        self.detach();
+    }
+}
+
+dedicated_add_obj!(
+    ZwlrForeignToplevelHandleV1,
+    ZwlrForeignToplevelHandleV1Id,
+    zwlr_foreign_toplevel_handles
+);
```
mahkoh: This doesn't need dedicated_add_obj.

src/ifs/zwlr_foreign_toplevel_manager_v1.rs
```diff
@@ -0,0 +1,168 @@
...
+    ZwlrForeignToplevelManagerV1Error
+);
+
+impl Global for ZwlrForeignToplevelManagerV1Global {
+    fn singleton(&self) -> bool {
+        true
+    }
+
+    fn version(&self) -> u32 {
+        1
```
mahkoh: 3

src/tree/toplevel.rs
```diff
@@ -94,6 +97,24 @@ impl<T: ToplevelNodeBase> ToplevelNode for T {
...
+                        .map(|k| k.1);
+                    if let Some(parent_handle) = parent_handle {
+                        handle.send_parent(parent_handle.clone());
+                        handle.send_done();
+                    }
+                }
+            }
+
+            handle.send_done();
+        }
```
mahkoh: This doesn't do anything since a toplevel advertised through this protocol can never be the parent of another toplevel.

src/tree/toplevel.rs
```diff
@@ -108,11 +129,20 @@ impl<T: ToplevelNodeBase> ToplevelNode for T {
...
         if prev_id != new_id {
             self.tl_workspace_output_changed();
+
+            let prev_output = prev.clone().map(|p| p.output.get());
+            let new_output = ws.output.get();
+
+            for handle in data.manager_handles.borrow().lock().values() {
+                handle.handle_output_changed(prev_output.clone(), new_output.clone());
+                handle.send_done();
+            }
```
mahkoh: This must go into `tl_workspace_output_changed`.

src/ifs/zwlr_foreign_toplevel_handle_v1.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    pub fn send_app_id(&self, app_id: &str) {
+        self.client.event(AppId {
+            self_id: self.id,
+            app_id,
+        });
+    }
+
+    pub fn send_state(&self, maximized: bool, minimized: bool, activated: bool, fullscreen: bool) {
```
mahkoh:
```suggestion
    pub fn send_state(&self, activated: bool, fullscreen: bool) {
```

disluckyguy:
Sorry commits are a bit of a mess now.

src/ifs/zwlr_foreign_toplevel_handle_v1.rs
disluckyguy: Done

src/ifs/zwlr_foreign_toplevel_handle_v1.rs
disluckyguy: Should be much simpler now

src/ifs/zwlr_foreign_toplevel_handle_v1.rs
disluckyguy: My bad, this implementation was a bit rushed, should be better now

src/tree/toplevel.rs
```diff
@@ -62,10 +66,22 @@ impl<T: ToplevelNodeBase> ToplevelNode for T {
         });
     }
 
-    fn tl_set_fullscreen(self: Rc<Self>, fullscreen: bool) {
+    fn tl_set_fullscreen(self: Rc<Self>, fullscreen: bool, output: Option<Rc<WlOutput>>) {
```
mahkoh: This should take an `Option<Rc<OutputNode>>`.

src/ifs/zwlr_foreign_toplevel_manager_v1.rs
```diff
@@ -0,0 +1,164 @@
...
+}
+
+impl ZwlrForeignToplevelManagerV1RequestHandler for ZwlrForeignToplevelManagerV1 {
+    type Error = ZwlrForeignToplevelManagerV1Error;
+
+    fn stop(&self, _req: Stop, _slf: &Rc<Self>) -> Result<(), Self::Error> {
+        self.detach();
+        self.send_finished();
+        Ok(())
+    }
```
mahkoh: This needs to destroy the object.

src/ifs/zwlr_foreign_toplevel_handle_v1.rs
```diff
@@ -0,0 +1,194 @@
...
+        Ok(())
+    }
+
+    fn set_rectangle(&self, _req: SetRectangle, _slf: &Rc<Self>) -> Result<(), Self::Error> {
+        Ok(())
+    }
+
+    fn set_fullscreen(&self, req: SetFullscreen, _slf: &Rc<Self>) -> Result<(), Self::Error> {
+        if let Some(toplevel) = self.toplevel.get() {
+            let output = self.client.objects.outputs.get(&req.output);
```
mahkoh: `self.client.lookup(req.output)`.

src/ifs/zwlr_foreign_toplevel_handle_v1.rs
```diff
@@ -0,0 +1,194 @@
...
+        Ok(())
+    }
+
+    fn unset_minimized(&self, _req: UnsetMinimized, _slf: &Rc<Self>) -> Result<(), Self::Error> {
+        Ok(())
+    }
+
+    fn activate(&self, _req: Activate, _slf: &Rc<Self>) -> Result<(), Self::Error> {
+        if let Some(toplevel) = self.toplevel.get() {
+            let seat = self.client.state.seat_queue.first();
```
mahkoh:
```suggestion
            let seat = self.client.state.seat_queue.last();
```

src/ifs/zwlr_foreign_toplevel_handle_v1.rs
```diff
@@ -0,0 +1,194 @@
...
+            if let Some(seat) = seat {
+                seat.focus_node(toplevel);
```
mahkoh:
```suggestion
                toplevel.node_do_focus(&seat, Direction::Unspecified);
```

src/ifs/zwlr_foreign_toplevel_handle_v1.rs
```diff
@@ -0,0 +1,194 @@
...
+
+    fn set_minimized(&self, _req: SetMinimized, _slf: &Rc<Self>) -> Result<(), Self::Error> {
+        Ok(())
+    }
+
+    fn unset_minimized(&self, _req: UnsetMinimized, _slf: &Rc<Self>) -> Result<(), Self::Error> {
+        Ok(())
+    }
+
+    fn activate(&self, _req: Activate, _slf: &Rc<Self>) -> Result<(), Self::Error> {
```
mahkoh: This needs to check that the toplevel is visible.

src/ifs/zwlr_foreign_toplevel_handle_v1.rs
```diff
@@ -0,0 +1,194 @@
...
+    }
+}
+
+impl ZwlrForeignToplevelHandleV1RequestHandler for ZwlrForeignToplevelHandleV1 {
+    type Error = ZwlrForeignToplevelHandleV1Error;
+
+    fn destroy(&self, _req: Destroy, _slf: &Rc<Self>) -> Result<(), Self::Error> {
+        self.detach();
+        Ok(())
+    }
```
mahkoh: This needs to destroy the object.

src/ifs/zwlr_foreign_toplevel_handle_v1.rs
disluckyguy: why not:
```rust
let seat = self.client.lookup(req.seat)
```

mahkoh: I just know that seat_queue.first is definitely wrong.

disluckyguy:
Currently some apps (tried ironbar and waybar) give an error when using the X11 backend (metal works fine), and that's the only issue left that I could think of. I'll probably recheck some functionality though

mahkoh:
I've made some significant changes. In the future, please do not use merges.

Please test this version.

mahkoh:
I believe set_rectangle is for animations when a window is minimized. It should not apply to tiling compositors.

src/tree/toplevel.rs
```diff
@@ -508,12 +554,72 @@ impl ToplevelData {
...
+        let opt = ToplevelOpt {
+            toplevel: Rc::downgrade(toplevel),
+            identifier: self.identifier.get(),
+        };
+        let handle = match manager.publish_toplevel(opt) {
+            None => return,
+            Some(handle) => handle,
+        };
+        handle.send_app_id(app_id);
+        handle.send_title(title);
```
disluckyguy: There should be checks on the title and app id, just to make sure they're not sent empty.

I'm keeping this review until I have time to do it

disluckyguy: Done

src/tree/toplevel.rs
```diff
@@ -508,12 +554,76 @@ impl ToplevelData {
...
+        let handle = match manager.publish_toplevel(opt) {
+            None => return,
+            Some(handle) => handle,
+        };
+        if !app_id.is_empty() {
+            handle.send_app_id(app_id);
+        }
+        if !title.is_empty() {
+            handle.send_title(title);
+        }
```
mahkoh, line 609: Why these conditions?

disluckyguy: In case the wm_class and title are empty/not sent it doesn't make sense to send it at the start

mahkoh: Is there any harm in sending it?

disluckyguy: if the client expects an empty id/title then there shouldn't be, I feel it's more correct to avoid that though, atleast when first sending the data.

Issue #332, opened 2024-12-14
Rust E0063 and E0308


apprehensions:
```
warning: use of deprecated type alias `uapi::c::time_t`: This type is changed to 64-bit in musl 1.2.0, we'll follow that change in the future release. See #1848 for more info.
   --> src/time.rs:105:59
    |
105 |         let secs = (rhs.as_nanos() / 1_000_000_000) as c::time_t;
    |                                                           ^^^^^^
    |
    = note: `#[warn(deprecated)]` on by default

error[E0063]: missing field `__pad1` in initializer of `cmsghdr`
  --> src/io_uring/ops/sendmsg.rs:47:27
   |
47 |                 let hdr = c::cmsghdr {
   |                           ^^^^^^^^^^ missing `__pad1`

error[E0308]: mismatched types
  --> src/io_uring/ops/sendmsg.rs:69:36
   |
69 |             st.msghdr.msg_iovlen = st.iovecs.len();
   |             --------------------   ^^^^^^^^^^^^^^^ expected `i32`, found `usize`
   |             |
   |             expected due to the type of this binding

error[E0308]: mismatched types
   --> src/udmabuf.rs:113:52
    |
113 |         let dmabuf = unsafe { ioctl(self.fd.raw(), UDMABUF_CREATE, &mut cmd) };
    |                               -----                ^^^^^^^^^^^^^^ expected `i32`, found `u64`
    |                               |
    |                               arguments to this function are incorrect
    |
note: function defined here
   --> /home/sewn/.local/share/cargo/registry/src/index.crates.io-6f17d22bba15001f/libc-0.2.153/src/unix/linux_like/linux/musl/mod.rs:850:12
    |
850 |     pub fn ioctl(fd: ::c_int, request: ::c_int, ...) -> ::c_int;
    |            ^^^^^
help: you can convert a `u64` to an `i32` and panic if the converted value doesn't fit
    |
113 |         let dmabuf = unsafe { ioctl(self.fd.raw(), UDMABUF_CREATE.try_into().unwrap(), &mut cmd) };
    |                                                                  ++++++++++++++++++++

error[E0308]: mismatched types
   --> src/video/dmabuf.rs:82:44
    |
82  |     let res = unsafe { ioctl(dmabuf.raw(), DMA_BUF_IOCTL_EXPORT_SYNC_FILE, &mut data) };
    |                        -----               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `i32`, found `u64`
[...]

PR #455, opened 2025-05-08
add musl support

elde-n:
Closes: #332


mahkoh:
Please create a module under utils that contains type aliases for ioctl numbers and the `iovecs_len` function.

mahkoh:
Thanks!

PR #469, opened 2025-05-12
input: add click method and middle button emulation

### LINKED ISSUE #467: ability to configure libinput click method and middle button emulation for touchpads
libinput has options to configure click method and middle button emulation.

https://wayland.freedesktop.org/libinput/doc/latest/configuration.html#click-method

https://wayland.freedesktop.org/libinput/doc/latest/configuration.html#middle-button-emulation

Would be nice to expose them in the config so they can be adjusted as needed.

kotarac:
closes #467 

Here's an attempt to cover #467. Let me know if I'm missing anything.

Testing it locally with a couple of different combinations of settings and everything seems to work as expected, incl. changing them in runtime.

libinput ref https://wayland.freedesktop.org/libinput/doc/latest/api/libinput_8h.html


jay-config/src/_private/ipc.rs
```diff
@@ -388,6 +388,14 @@ pub enum ClientMessage<'a> {
...
         enabled: bool,
     },
+    SetClickMethod {
+        device: InputDevice,
+        method: ClickMethod,
+    },
+    SetMiddleButtonEmulationEnabled {
+        device: InputDevice,
+        enabled: bool,
+    },
```
mahkoh: New messages must be added at the end.

jay-config/src/input/acceleration.rs
```diff
@@ -6,7 +6,7 @@ use serde::{Deserialize, Serialize};
 
 /// The acceleration profile of a device.
 #[derive(Serialize, Deserialize, Copy, Clone, Debug, Hash, Eq, PartialEq)]
-pub struct AccelProfile(pub u32);
+pub struct AccelProfile(pub i32);
```
mahkoh: The public API cannot be changed.

wire/jay_input.txt
```diff
@@ -151,6 +161,9 @@ event input_device {
     left_handed: u32,
     natural_scrolling_available: u32,
     natural_scrolling_enabled: u32,
+    click_method: i32,
+    middle_button_emulation_available: u32,
+    middle_button_emulation_enabled: u32,
```
mahkoh: Existing events cannot be modified.

wire/jay_input.txt
```diff
@@ -64,6 +64,16 @@ request set_natural_scrolling {
...
 
+request set_click_method {
+    id: u32,
+    method: i32,
+}
+
+request set_middle_button_emulation {
+    id: u32,
+    enabled: u32,
+}
```
mahkoh: New requests must be added below existing requets.

toml-config/src/config/extractor.rs
```diff
@@ -277,3 +277,5 @@ tuples!(0:T0,1:T1,2:T2,3:T3,4:T4,5:T5,6:T6,);
 tuples!(0:T0,1:T1,2:T2,3:T3,4:T4,5:T5,6:T6,7:T7,);
 tuples!(0:T0,1:T1,2:T2,3:T3,4:T4,5:T5,6:T6,7:T7,8:T8,);
 tuples!(0:T0,1:T1,2:T2,3:T3,4:T4,5:T5,6:T6,7:T7,8:T8,9:T9,);
+tuples!(0:T0,1:T1,2:T2,3:T3,4:T4,5:T5,6:T6,7:T7,8:T8,9:T9,10:T10,);
+tuples!(0:T0,1:T1,2:T2,3:T3,4:T4,5:T5,6:T6,7:T7,8:T8,9:T9,10:T10,11:T11,);
```
mahkoh: These should not be necessary.

kotarac: without it

```
error[E0277]: the trait bound `((Opt<impl FnOnce(&mut Extractor<'a>) -> Result<Spanned<&str>, Spanned<...>>>, ..., ..., ..., ..., ..., ..., ..., ..., ..., ..., ...), ...): Extractable<'_>` is not satisfied
   --> toml-config/src/config/parsers/input.rs:94:25
    |
94  |           ) = ext.extract((
    |  _________________-------_^
    | |                 |
    | |                 required by a bound introduced by this call
95  | |             (
96  | |                 opt(str("tag")),
97  | |                 val("match"),
...   |
119 | |             ),
120 | |         ))?;
    | |_________^ the trait `Extractable<'_>` is not implemented for `((Opt<impl FnOnce(&mut Extractor<'a>) -> Result<Spanned<&str>, Spanned<ExtractorError>>>, ..., ..., ..., ..., ..., ..., ..., ..., ..., ..., ...), ...)`
    |
    = help: the following other types implement trait `Extractable<'v>`:
              (T0, T1)
              (T0, T1, T2)
              (T0, T1, T2, T3)
              (T0, T1, T2, T3, T4)
              (T0, T1, T2, T3, T4, T5)
              (T0, T1, T2, T3, T4, T5, T6)
              (T0, T1, T2, T3, T4, T5, T6, T7)
              (T0, T1, T2, T3, T4, T5, T6, T7, T8)
            and 2 others
note: required by a bound in `Extractor::<'v>::extract`
   --> toml-config/src/config/extractor.rs:50:23
    |
50  |     pub fn extract<E: Extractable<'v>, U>(&mut self, e: E) -> Result<E::Output, Spanned<U>>
    |                       ^^^^^^^^^^^^^^^ required by this bound in `Extractor::<'v>::extract`
    = note: the full name for the type has been written to '/home/kotarac/proj/jay/target/debug/deps/jay_toml_config.long-type-597141285022403644.txt'
    = note: consider using `--verbose` to print the full type name to the console
```

mahkoh: You have to add new fields at the end.

wire/jay_input.txt
kotarac: hmm, how should adding these work then?

mahkoh: Same as `CalibrationMatrix`.

kotarac: Any other preferences?

I'm thinking there might be other fields like this added in the future, so it probably makes more sense to create separate events for each setting. What do you think?

mahkoh: Fine with me.

kotarac:
@mahkoh can you take another look?

Other than discussed changes, we might want to consider checking the output `libinput_device_config_click_get_methods` for the list of available click methods, but I didn't feel like it was worth it being that precise for this.

mahkoh:
I've made the following changes:

- Guarded click methods behind the device advertising any.
- Added the click method to fetch_effective.
- Removed the _available fields from jay_input events. Availability is implied by the message being sent. These fields only exist for the old event because everything is sent in one event.

mahkoh:
Thanks!

Issue #242, opened 2024-08-01
Multiple output configuration profiles for docked laptops


kennylevinsen:
Laptops may be docked to more than a single setup, e.g. a home and work setup. These may require different output configurations, such as one having the laptop screen to the right of external screens, another having to the left or below external screens.

Jey currently only support per-output configuration with positive coordinates, and does not support dynamic configuration protocols that would allow kanshi and its profiles to work.

There are three main ways for a user to handle this:
1. Dynamic configuration daemons like kanshi (or equivalent built-in functionality) that allow per-setup profiles to be applied.
2. Place the laptop display at 0,0 and rely on both positive and negative positioning to place other displays statically around it.
3. Place the laptop display arbitrarily deep into the global coordinate space so that the top-left-most display in any configuration is at 0,0, with other configurations leaving a top/left gap

Neither 1 or 2 is possible right now.

mahkoh:
>Jey currently only support per-output configuration with positive coordinates

This is deliberate bacause some downstream components cannot handle negative positions. I believe this affects both X and libei. But the user can just as well use 1000000x1000000 as their origin.

>and does not support dynamic configuration protocols that would allow kanshi and its profiles to work.

There are two ways to handle this right now. You could use a shortcut to reconfigure your outputs via the configure-output action https://github.com/mahkoh/jay/blob/master/toml-spec/spec/spec.generated.md#action. Or you can write a config.so that handles this automatically https://docs.rs/jay-config/latest/jay_config/.

>Place the laptop display arbitrarily deep into the global coordinate space so that the top-left-most display in any configuration is at 0,0

There is no requirement for 0x0 to be next to an output.

kennylevinsen:
Yeah we allow negative coordinates in sway, but it upsets Xwayland indeed.

Building a config.so would be a possibility for this use-case, but coding the logic to configure outputs certainly seems like higher effort than e.g. kanshi profiles, especially when adjustments are needed. The convenience of loading such profiles could of course be coded within the config.so...

mahkoh:
Adding support for the wlr protocol would be possible. I would prefer an upstream protocol but I should not ask for the impossible.

But I see a few issues.

The implementation would not be trivial because I currently don't do grouped device modifications. Instead individual requests are just passed into the backend and it either applies them or not.

My adaptive sync configuration is more sophisticated than a simple on off switch.

[...]

PR #477, opened 2025-06-03
implement wlr_output_management_unstable_v1

disluckyguy:
This will hopefully help jay to work better with wlroots program like wlr-randr.

### DIFF
--- a/src/backends/metal/video.rs
+++ b/src/backends/metal/video.rs
@@ -859,6 +859,11 @@ impl Connector for MetalConnector {
         if self.enabled.replace(enabled) != enabled {
             if self.display.borrow_mut().connection == ConnectorStatus::Connected {
                 if let Some(dev) = self.backend.device_holder.drm_devices.get(&self.dev.devnum) {
+                    if let Some(node) = self.state.root.outputs.get(&self.connector_id) {
+                        for head in node.zwlr_output_heads.lock().values() {
+                            head.handle_enabled_change(enabled);
+                        }
+                    }
                     if let Err(e) = self.backend.handle_drm_change_(&dev, true) {
                         dev.unprocessed_change.set(true);
                         log::error!("Could not dis/enable connector: {}", ErrorFmt(e));
--- a/src/backends/x.rs
+++ b/src/backends/x.rs
@@ -965,7 +965,7 @@ impl XBackend {
             output.events.push(ConnectorEvent::ModeChanged(Mode {
                 width,
                 height,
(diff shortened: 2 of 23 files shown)


mahkoh:
I don't believe the wlr output management protocol is very well designed due to its monolithic structure. I've gone into a bit more details here: #242.

I've been thinking about creating a new set of protocols for this purpose for some time.

mahkoh:
Actually you might want to hold off for now. I have another PR that should clarify some things.

disluckyguy:
I tested with kansh, wlr-randr, and a bit with my one client, a few thing to point out:
- We should probably include the connector name instead of the serial number in the description or at least include it like other compositors
- We should avoid non-adjacent outputs when scaling(see #493 ) and positioning because of their clunky behavior, maybe in another mr if you want.
- a bit unrelated, but the documentation for the protocol states:
> If the compositor implements xdg-output and this head is enabled, the xdg_output.description must report the same description.

description is not implemented for xdg-output in jay, I think we should implement it there

Other than that it works well, thank you for the rebase, and sorry that my initial implementaion was...not the best

mahkoh:
>We should probably include the connector name instead of the serial number in the description or at least include it like other compositors

It seems that sway includes both the connector name and the serial number in its description. But I do not believe that including the connector name is a good idea because connector names are not stable across reboots.

>We should avoid non-adjacent outputs when scaling(see https://github.com/mahkoh/jay/issues/493 ) and positioning because of their clunky behavior, maybe in another mr if you want.

I believe it is the job of the client configuring the outputs to correctly position them.

>description is not implemented for xdg-output in jay, I think we should implement it there

Ok

src/ifs/wlr_output_manager/zwlr_output_configuration_head.rs
```diff
@@ -44,6 +45,9 @@ impl ZwlrOutputConfigurationHeadV1RequestHandler for ZwlrOutputConfigurationHead
             return Err(ZwlrOutputConfigurationHeadV1Error::AlreadySet);
         }
         let mode = self.client.lookup(req.mode)?;
+        if !self.head.modes.values().any(|m| m.id == req.mode) {
+            return Err(ZwlrOutputConfigurationHeadV1Error::InvalidMode);
+        }
```
mahkoh: I don't believe this is necessary. But if you want to implement this then create a new `linear_id` identifying a `ZwlrOutputHeadV1`.

PR #489, opened 2025-06-21
Set top layer surfaces below fullscreen windows

### LINKED ISSUE #476: layer shell windows in "Top" layer on top of fullscreen windows
From my testing in sway and labwc, this behavior only happens on "Overlay" layer shells, which makes sense since it will allow overlays like launchers to work but cover for example a bar or dock, which commonly use the "Top" layer.

disluckyguy:
closes #476


mahkoh:
Thanks!

PR #497, opened 2025-07-01
Focus improvements

disluckyguy:
This pr does two things:
- implements a focus queue that is used so that switching workspaces it focuses the last focused toplevel.
- wlr_foreign_toplevel_handles's focus request show the toplevels workspace

I can make them in separate pull requests if you want


src/ifs/zwlr_foreign_toplevel_handle_v1.rs
```diff
@@ -62,11 +62,14 @@ impl ZwlrForeignToplevelHandleV1RequestHandler for ZwlrForeignToplevelHandleV1 {
 
     fn activate(&self, req: Activate, _slf: &Rc<Self>) -> Result<(), Self::Error> {
         if let Some(toplevel) = self.toplevel.get() {
-            if !toplevel.node_visible() {
-                return Ok(());
-            }
```
mahkoh: This check is still required before calling node_do_focus.

src/state.rs
```diff
@@ -814,6 +814,7 @@ impl State {
     }
 
     pub fn float_map_ws(&self) -> Rc<WorkspaceNode> {
+        println!("mapping float");
```
mahkoh: Remove this

src/tree/toplevel.rs
```diff
@@ -417,6 +429,22 @@ impl ToplevelData {
                 handle.send_state(active_new, self.is_fullscreen.get());
                 handle.send_done();
             }
+            if active_new {
+                if let Some(workspace) = self.workspace.get() {
+                    let link = workspace
+                        .focus_history
+                        .iter()
+                        .find(|lf| lf.node_id() == tl.node_id());
```
mahkoh: Use self.focus_link insetad.

src/tree/toplevel.rs
```diff
@@ -492,6 +520,14 @@ impl ToplevelData {
...
         self.workspace.take();
         self.seat_state.destroy_node(node);
+        self.detach_focus_link();
+    }
+
+    pub fn detach_focus_link(&self) {
+        let link = self.focus_link.take();
+        if let Some(link) = link {
+            link.detach();
+        }
```
mahkoh: This is unnecessary.

src/tree/toplevel.rs
disluckyguy: I feel it would be a bit awkward to switch to workspace without a focused window or the wrong focused window.

mahkoh: I don't understand what you mean by this comment. The change you made is what I meant.

disluckyguy: Sorry, I mean that if a toplevel is not visible the workspace will show but either the wrong window will be focus or the none would be focused. 
I feel like that would be strange for the user, or would this just never happen?

disluckyguy: Oh, I was replying to the wrong review. I'm so sorry. This one should be fine

disluckyguy:
any work remaining to merge?

mahkoh:
I believe this should be implemented by #521 and #523. Please test the current master.

src/ifs/zwlr_foreign_toplevel_handle_v1.rs
```diff
@@ -62,11 +62,23 @@ impl ZwlrForeignToplevelHandleV1RequestHandler for ZwlrForeignToplevelHandleV1 {
...
+
             if !toplevel.node_visible() {
                 return Ok(());
             }
-            let seat = self.client.lookup(req.seat)?;
-            toplevel.node_do_focus(&seat.global, Direction::Unspecified);
+
+            toplevel
+                .clone()
+                .node_do_focus(&seat.global, Direction::Unspecified);
```
mahkoh, line 81: This change is unnecessary.

src/tree/toplevel.rs
```diff
@@ -44,7 +45,7 @@ use {
             JayToplevelId, ZwlrForeignToplevelHandleV1Id,
         },
     },
-    jay_config::{window, window::WindowType},
+    jay_config::window::{self, WindowType},
```
mahkoh, line 48: This change is unnecessary.

mahkoh:
>Unrelated, but do you have any feedback on my recent mrs?

I've added two examples above of changes that should not be part of this PR.

Some more things:

- Please use rebasing to amend your previous commits instead of pushing commits like "resolve reviews"
- Please use commit subject lines that follow the general style of existing commits

I don't remember any other issues that aren't just due to unfamiliarity with the existing code.

PR #498, opened 2025-07-01
all: use let chains

mahkoh:


mahkoh:
In RustRover I used the "Find in Files" functionality to find all occurrences of "if let" and then when through them manually.

PR #506, opened 2025-07-17
tree: restack containing float when node is activated

mahkoh:


mahkoh:
Should be fixed by https://github.com/mahkoh/jay/pull/509

PR #509, opened 2025-07-17
tablet: restack float on down instead of button

mahkoh:


disluckyguy:
It indeed works for my stylus, thank you

PR #539, opened 2025-07-24
build: add nix flake

mahkoh:
Based on https://github.com/YaLTeR/niri/blob/main/flake.nix and https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/ja/jay/package.nix.


mahkoh:
Blocked by https://github.com/emersion/hottub/issues/4

Richochet1:
You may also want to add a session file, so that you can use `jay` with display managers

Issue #519, opened 2025-07-18
workspace bar enhancements


kotarac:
Couple of wishlist items for the workspace display in the bar.

1. sort workspaces by name / number
Assuming you have workspaces 1, 2, ..., 0, jay does not sort the currently active ones, so you can regularly get something like in the screenshot below.
<img width="248" height="36" alt="Image" src="https://github.com/user-attachments/assets/798182b1-974b-4eea-bad8-e456e17ee3f3" />

2. automatically hide the workspace indicator when it has no more children
When you don't have a workspace focused and the last window on it closes, the workspace indicator in the bar doesn't hide.
After going to the, then empty, workspace and back to some other, the indicator then disappear.

PR #541, opened 2025-07-25
tree: skip empty inactive workspaces

kotarac:
Handles point 2 from #519.

Not sure how it should behave depending on `ws.attention_requests`. Can those happen if the workspace is empty?

### DIFF
--- a/src/tree/output.rs
+++ b/src/tree/output.rs
@@ -590,6 +590,9 @@ impl OutputNode {
         let output_width = non_exclusive_rect.width();
         rd.underline = Rect::new_sized(0, th, output_width, 1).unwrap();
         for ws in self.workspaces.iter() {
+            if ws.is_empty() && !ws.has_capture.get() && Some(ws.id) != active_id {
+                continue;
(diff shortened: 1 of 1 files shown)


mahkoh:
That's more complicated. You can manually drag workspaces around so you'd have to think about how these two things interact.

kotarac:
> That's more complicated. You can manually drag workspaces around so you'd have to think about how these two things interact.

I was thinking making it configurable and if sorting is enabled having dragging disabled.

> Hide differs from destroy in that reopening the workspace opens it on the same output it was previously on.

Is hide a new concept or is there precedent in jay?

mahkoh:
>I was thinking making it configurable and if sorting is enabled having dragging disabled.

That sounds reasonable.

>Is hide a new concept or is there precedent in jay?

It's what your code does at the moment.

kotarac:
Closing this for now. Will take a stab at it according to https://github.com/mahkoh/jay/pull/541#issuecomment-3120206447 sometime.

PR #542, opened 2025-07-26
config: workspace display order

kotarac:
Point 1 from #519 done as we talked about here https://github.com/mahkoh/jay/pull/541#issuecomment-3120275757.

- considered natural sorting but decided it's probably not worth it
- maintains a sorted linked list of workspaces if workspace-display-order is set to sorted, handles workspace creation, drag, move-to-output
- O(N) search to find an appropriate place to insert a workspace, but it's only on user actions and it's unlikely there's ever going to be many of them, so should be fine

### DIFF
--- a/jay-config/src/lib.rs
+++ b/jay-config/src/lib.rs
@@ -45,6 +45,7 @@
 
 #[expect(unused_imports)]
 use crate::input::Seat;
+
 use {
     crate::{
         _private::ipc::WorkspaceSource, keyboard::ModifiedKeySym, video::Connector, window::Window,
@@ -73,6 +74,7 @@ pub mod theme;
 pub mod timer;
 pub mod video;
 pub mod window;
+pub mod workspace;
 pub mod xwayland;
 
 /// A planar direction.
--- a/src/config/handler.rs
+++ b/src/config/handler.rs
@@ -74,6 +74,7 @@ use {
             Transform, VrrMode as ConfigVrrMode,
         },
         window::{TileState, Window, WindowMatcher},
+        workspace::WorkspaceDisplayOrder,
         xwayland::XScalingMode,
     },
     kbvm::Keycode,
@@ -1350,6 +1351,13 @@ impl ConfigProxyHandler {
             if let Some(float) = stacked.deref().clone().node_into_float() {
                 float.schedule_render_titles();
             }
+        }
+    }
+
+    fn handle_set_workspace_display_order(&self, order: WorkspaceDisplayOrder) {
+        self.state.workspace_display_order.set(order);
+        for output in self.state.root.outputs.lock().values() {
+            output.handle_workspace_display_order_update();
         }
     }
 
@@ -3097,6 +3105,9 @@ impl ConfigProxyHandler {
             }
             ClientMessage::SetMiddleClickPasteEnabled { enabled } => {
                 self.handle_set_middle_click_paste_enabled(enabled)
+            }
+            ClientMessage::SetWorkspaceDisplayOrder { order } => {
+                self.handle_set_workspace_display_order(order)
             }
             ClientMessage::SeatCreateMark { seat, kc } => self
                 .handle_seat_create_mark(seat, kc)
--- a/src/tree/output.rs
+++ b/src/tree/output.rs
@@ -48,17 +48,27 @@ use {
             WorkspaceDragDestination, WorkspaceNode, WorkspaceNodeId, walker::NodeVisitor,
         },
         utils::{
-            asyncevent::AsyncEvent, bitflags::BitflagsExt, clonecell::CloneCell,
-            copyhashmap::CopyHashMap, errorfmt::ErrorFmt, event_listener::EventSource,
-            hash_map_ext::HashMapExt, linkedlist::LinkedList, on_drop_event::OnDropEvent,
-            scroller::Scroller, transform_ext::TransformExt,
+            asyncevent::AsyncEvent,
+            bitflags::BitflagsExt,
+            clonecell::CloneCell,
+            copyhashmap::CopyHashMap,
+            errorfmt::ErrorFmt,
+            event_listener::EventSource,
+            hash_map_ext::HashMapExt,
+            linkedlist::{LinkedList, NodeRef},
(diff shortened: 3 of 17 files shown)


jay-config/src/lib.rs
```diff
@@ -353,3 +356,10 @@ pub fn on_unload(f: impl FnOnce() + 'static) {
 pub fn set_middle_click_paste_enabled(enabled: bool) {
     get!().set_middle_click_paste_enabled(enabled);
 }
+
+/// Sets how workspaces should be ordered in the UI.
+///
+/// The default is `WorkspaceDisplayOrder::Manual`.
+pub fn set_workspace_display_order(order: WorkspaceDisplayOrder) {
+    get!().set_workspace_display_order(order);
+}
```
mahkoh: Move this to the workspace module.

src/tree/output.rs
```diff
@@ -109,6 +112,7 @@ pub struct OutputNode {
     pub ext_workspace_groups: CopyHashMap<WorkspaceManagerId, Rc<ExtWorkspaceGroupHandleV1>>,
     pub pinned: LinkedList<Rc<dyn PinnedNode>>,
     pub tearing: Cell<bool>,
+    pub workspace_display_order: Rc<Cell<WorkspaceDisplayOrder>>,
```
mahkoh: This field is unnecessary.

src/tree/output.rs
```diff
@@ -1274,6 +1294,9 @@ impl OutputNode {
         x_abs: i32,
         y_abs: i32,
     ) -> Option<WorkspaceDragDestination> {
+        if self.workspace_display_order.get() == WorkspaceDisplayOrder::Sorted {
+            return None;
+        }
```
mahkoh: It should still be possible to drag workspaces between outputs.

toml-config/src/config/parsers/config.rs
```diff
@@ -528,6 +533,18 @@ impl Parser for ConfigParser<'_> {
...
+                    Ok(v) => Some(v),
+                    Err(e) => {
+                        log::warn!(
+                            "Could not parse the workspace display order: {}",
+                            self.0.error(e)
+                        );
+                        None
+                    }
+                }
+            }),
```
mahkoh: Move this out of the struct construction.

mahkoh:
I think it would be better if the workspaces were sorted in the linked list.

src/tree/output.rs
kotarac: Do you consider this a hard requirement? I'd personally prefer this not being possible, or ideally configurable as well.

kotarac:
> I think it would be better if the workspaces were sorted in the linked list.

You mean keeping the list of workspaces sorted at all times when it's updated?

mahkoh:
Yes.

src/tree/output.rs
mahkoh: Yes, that is a hard requirement. I think it should be easy to implement. Something like

```rust
        if self.workspace_display_order.get() == WorkspaceDisplayOrder::Sorted {
            if the workspace is already on this output return none
            return Some(WorkspaceDragDestination {
                highlight: Rect::new_sized(
                    rect.x1(),
                    rect.y1(),
                    rect.width(),
                    th,
                )?,
                output: self.clone(),
                before: None,
            });
        }
```

kotarac: > I think it should be easy to implement.

Agreed, just don't have anything to reliably test it with right now. :slightly_smiling_face:

src/config/handler.rs
```diff
@@ -1353,6 +1357,13 @@ impl ConfigProxyHandler {
         }
     }
 
+    fn handle_set_workspace_display_order(&self, order: WorkspaceDisplayOrder) {
+        self.state.workspace_display_order.set(order);
+        for output in self.state.root.outputs.lock().values() {
+            output.schedule_update_render_data();
+        }
```
mahkoh: The existing workspaces need to be sorted here if necessary.

src/tree/output.rs
```diff
@@ -712,6 +722,17 @@ impl OutputNode {
         true
     }
 
+    pub fn find_workspace_insertion_point(&self, name: &str) -> Option<NodeRef<Rc<WorkspaceNode>>> {
+        if self.state.workspace_display_order.get() == WorkspaceDisplayOrder::Sorted {
+            for existing_ws in self.workspaces.iter() {
+                if name.cmp(&existing_ws.name).is_lt() {
```
mahkoh: Why not just `name < existing_ws.name`?

src/config/handler.rs
```diff
@@ -1073,11 +1074,14 @@ impl ConfigProxyHandler {
...
+            .find_workspace_insertion_point(&ws.name)
+            .map(|nr| nr.deref().clone());
         let config = WsMoveConfig {
             make_visible_always: false,
             make_visible_if_empty: true,
             source_is_destroyed: false,
-            before: None,
+            before,
         };
         move_ws_to_output(&link, &output, config);
```
mahkoh: How about letting `move_ws_to_output` handle find the insertion point?

kotarac: Sure you want that changed? Couple of other things also depend on `move_ws_to_output` with `WsMoveConfig.before`.

mahkoh: >Sure you want that changed?

Yes

kotarac:
Addressed those points. Squashed the commits, noticed in other PRs you prefer it that way and I've done it like that before as well.

src/tree/output.rs
```diff
@@ -1285,6 +1323,29 @@ impl OutputNode {
...
+                .lock()
+                .values()
+                .find(|ws| ws.id == source)
+            {
+                Some(ws) => ws.name.clone(),
+                None => return None,
+            };
+            let before = self
+                .find_workspace_insertion_point(&source_name)
+                .map(|nr| nr.deref().clone());
```
mahkoh: I think this can all be removed now.

kotarac: Oh right. Just now realized that was the whole point of moving that logic to `move_ws_to_output`.

Done. `before: None` there should be ok now I believe.

PR #548, opened 2025-07-27
layershell: update tree on position change

### LINKED ISSUE #546: Layer shell windows don't update their positions
jay doesn't update a layer shell window if it changes its position without changing its size (e.g a bottom bar setting a bottom margin, or setting the anchor from the top to the bottom)

This seems to be due to this condition
```rust
if self.surface.extents.get().size() != self.pos.get().size()
       {
            self.compute_position();
        }
```
so either removing this condition or recalculating when the position changes without the size would work

disluckyguy:
fixes #546

### DIFF
--- a/src/ifs/wl_surface/zwlr_layer_surface_v1.rs
+++ b/src/ifs/wl_surface/zwlr_layer_surface_v1.rs
@@ -71,6 +71,7 @@ pub struct ZwlrLayerSurfaceV1 {
     exclusive_edge: Cell<Option<u32>>,
     exclusive_size: Cell<ExclusiveSize>,
     popups: CopyHashMap<XdgPopupId, Rc<Popup>>,
+    need_position_update: Cell<bool>,
 }
 
 #[derive(Copy, Clone, Debug, Default, PartialEq, Eq)]
@@ -177,6 +178,7 @@ impl ZwlrLayerSurfaceV1 {
             exclusive_edge: Default::default(),
             exclusive_size: Default::default(),
             popups: Default::default(),
+            need_position_update: Default::default(),
         }
     }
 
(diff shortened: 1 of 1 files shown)


mahkoh:
I don't think this is the correct solution. Instead you should check in pre_commit if any of the input to compute_position have changed and then set a flag.

disluckyguy:
any reason we need to use flags? I don't think what the changes are matters so a boolean would be enough

mahkoh:
That's what I meant.

disluckyguy:
I've added some checks to the margin to avoid unnecessary changes, if you want to check if the margin actually changes the position (a change to right margin of a bottom bar won't change it's position for example) I can also do that.

the current state looks closer to what you want though

mahkoh:
Please test this version.

src/ifs/wl_surface/zwlr_layer_surface_v1.rs
```diff
@@ -398,13 +400,19 @@ impl ZwlrLayerSurfaceV1 {
...
         if let Some(anchor) = pending.anchor.take() {
-            self.anchor.set(anchor);
+            if self.anchor.replace(anchor) != anchor {
+                self.need_position_update.set(true);
+            }
         }
         if let Some(ez) = pending.exclusive_zone.take() {
-            self.exclusive_zone.set(ez);
+            if self.exclusive_zone.replace(ez) != ez {
+                self.need_position_update.set(true);
```
disluckyguy, line 409: are you sure the exclusive zone can update the position?

src/ifs/wl_surface/zwlr_layer_surface_v1.rs
```diff
@@ -398,13 +400,19 @@ impl ZwlrLayerSurfaceV1 {
...
+            }
         }
         if let Some(margin) = pending.margin.take() {
-            self.margin.set(margin);
+            if self.margin.replace(margin) != margin {
+                self.need_position_update.set(true);
```
disluckyguy, line 414: Unanchored edges are not affected by margins, which is why I had the conditions previously.

disluckyguy (commented):
Besides these, it works, thank you.

src/ifs/wl_surface/zwlr_layer_surface_v1.rs
mahkoh: Yes

src/ifs/wl_surface/zwlr_layer_surface_v1.rs
mahkoh: That's a bit too fine-grained for me.

PR #552, opened 2025-07-27
layershell: properly set margins for size 0 layershell surfaces

### LINKED ISSUE #550: Margins not set properly for layershell windows
If the layershell has opposite edges anchored (TOP | BOTTOM and LEFT | RIGHT) the margins for these edges are added to one side (right for the former, bottom for the latter).

this video explains this

https://github.com/user-attachments/assets/f29aa178-d484-41e9-b632-bab832b5b32a

disluckyguy:
fixes #550 
This is similar to how it works on sway


mahkoh:
The protocol says that a surface that is anchored on two opposite edges will be centered.

disluckyguy:
I'll test this a bit more to check the behavior

mahkoh:
Have you tested the case where the client doesn't call set_size at all and then attaches a smaller buffer?

disluckyguy:
I'm sorry, I'm not really sure how to test this, my current testing client uses iced_layershell, I can try to make a client with raw wayland and wgpu but it will take quite some time. Will this really affect this mr? it mostly deals with margins, not setting the size will have the same behavior as before

to be more specific, here's what I tested:
- margins with orthogonal edge + opposite edges, both with a specific size and a width/height of 0

mahkoh:
If the client doesn't set a size or margin, your code puts them at the edge of the screen.

disluckyguy:
I could set the size to an option and only calculate if a size is set by the client

mahkoh:
No, setting the size to 0 explicitly should be the same as not setting the size at all. compute_position must not use this value.

disluckyguy:
I'm sorry, I still don't get how that's wrong, as the documentation states
> You must set your anchor to opposite edges in the dimensions you omit; not doing so is a protocol error.

So the only case where not setting a size (so both values are 0) is not a protocol error is when all anchors are set. in this case, the surface would fill the entire screen, so setting it to the edge won't matter. This will also allow that surface to set its margins properly. This is what happens in sway at least

mahkoh:
>in this case, the surface would fill the entire screen

It doesn't.

>If the client picks a smaller size and is anchored to two opposite anchors (e.g. 'top' and 'bottom'), the surface will be centered on this axis.

disluckyguy:
This is the closest I could get to what you want, the current problem is that it "shakes" when being set to some margins. I'll try to look into this

disluckyguy:
I added the checks because it seems the behavior I described only happenes when the size is set to 0.
For example, setting the size to the monitor width(so the full width, like what the compositor usually does with a width of 0) causes sway to ignore the margins, like it does with any set width, and unlike when the width is 0

PR #638, opened 2025-10-09
Remove titlebar separator if title_height is 0.

### LINKED ISSUE #619: One Pixel Offset
After applying these settings

```toml
show-bar = false

[theme]
border-width = 0
title-height = 0
```

There is a one pixel offset at the top of the screen.  The offset becomes apparent when toggling full screen for an application.

Config: [config.toml.txt](https://github.com/user-attachments/files/22458546/config.toml.txt)

sday-ssc:
If title_height is 0, the 1 pixel separator between titlebars and windows is no longer added.
Fixes #619


jay-config/src/_private/ipc.rs
```diff
@@ -727,6 +727,10 @@ pub enum ClientMessage<'a> {
         show: bool,
     },
     GetShowBar,
+    SetShowTitles {
+        show: bool,
+    },
+    GetShowTitles,
```

toml-config/src/config/parsers/action.rs
```diff
@@ -146,6 +146,9 @@ impl ActionParser<'_> {
             "show-bar" => ShowBar(true),
             "hide-bar" => ShowBar(false),
             "toggle-bar" => ToggleBar,
+            "show-title-bar" => ShowTitles(true),
+            "hide-title-bar" => ShowTitles(false),
+            "toggle-title-bar" => ToggleTitles,
```
mahkoh: Drop the -bar.

jay-config/src/_private/ipc.rs
sday-ssc: I'm not following on this. Do you mean the GetShowTitles variant in Response?

mahkoh: These enum variants must be moved to the bottom of the enum.

toml-spec/spec/spec.yaml
```diff
@@ -1157,14 +1163,14 @@ DrmDeviceMatch:
...
             This is useful if you have multiple copies of the same device installed so that
             the PCI numbers are not unique.
-            
+
             The values are usually stable unless you re-configure your hardware.
-            
+
             - Example:
-            
+
```
mahkoh: Please revert these changes.

sday-ssc:
Seems to work as expected!

PR #665, opened 2025-11-27
Add mouse-refocus option to center cursor on focused window

ArthurHeymans:
Implement automatic cursor positioning to the center of the active
window when focus changes via keyboard commands (move-left, focus-right,
show-workspace, etc.). This feature is enabled by default but can be
disabled in config.toml with `mouse-refocus = false`.

### DIFF
--- a/jay-config/src/input.rs
+++ b/jay-config/src/input.rs
@@ -638,6 +638,14 @@ impl Seat {
     pub fn enable_unicode_input(self) {
         get!().seat_enable_unicode_input(self);
     }
+
+    /// Sets whether the cursor should automatically move to the center of a window
+    /// when focus changes via keyboard commands (move-left, focus-right, show-workspace, etc.).
+    ///
+    /// Default: false.
+    pub fn set_mouse_refocus(self, enabled: bool) {
+        get!().seat_set_mouse_refocus(self, enabled)
+    }
 }
 
 /// A focus-follows-mouse mode.
--- a/src/ifs/wl_seat.rs
+++ b/src/ifs/wl_seat.rs
@@ -246,6 +246,7 @@ pub struct WlSeatGlobal {
     modifiers_forward: EventSource<dyn LedsListener>,
     simple_im: CloneCell<Option<Rc<SimpleIm>>>,
     simple_im_enabled: Cell<bool>,
+    pub mouse_refocus: Cell<bool>,
 }
 
 #[derive(Copy, Clone)]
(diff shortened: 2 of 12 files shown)


src/ifs/wl_seat.rs
```diff
@@ -745,6 +750,26 @@ impl WlSeatGlobal {
...
+        let Some(tl) = self.keyboard_node.get().node_toplevel() else {
+            return;
+        };
+        // Get the window's absolute global position and size
+        let geom = tl.node_absolute_position();
+        // Calculate center position in global coordinates
+        let center_x = Fixed::from_int((geom.x1() + geom.x2()) / 2);
+        let center_y = Fixed::from_int((geom.y1() + geom.y2()) / 2);
+        // Move the cursor to the center
+        self.pointer_cursor.set_position(center_x, center_y);
```
mahkoh:
```suggestion
        let (x, y) = tl.node_absolute_position().center();
        self.state.position_hint_requests.push(PositionHintRequest {
            seat: self.clone(),
            client_id: None,
            old_pos: self.pointer_cursor.position(),
            new_pos: (Fixed::from_int(x), Fixed::from_int(y)),
        });
```

jay-config/src/input.rs
```diff
@@ -638,6 +638,14 @@ impl Seat {
     pub fn enable_unicode_input(self) {
         get!().seat_enable_unicode_input(self);
     }
+
+    /// Sets whether the cursor should automatically move to the center of a window
+    /// when focus changes via keyboard commands (move-left, focus-right, show-workspace, etc.).
+    ///
+    /// Default: true.
```
mahkoh:
```suggestion
    /// Default: false.
```

toml-config/src/lib.rs
```diff
@@ -1617,6 +1617,8 @@ fn load_config(initial_load: bool, auto_reload: bool, persistent: &Rc<Persistent
             persistent.seat.set_simple_im_enabled(enabled);
         }
     }
+    let mouse_refocus = config.mouse_refocus.unwrap_or(true);
+    persistent.seat.set_mouse_refocus(mouse_refocus);
```
mahkoh:
```suggestion
    if let Some(mouse_refocus) = config.mouse_refocus {
        persistent.seat.set_mouse_refocus(mouse_refocus);
    }
```

src/ifs/wl_seat.rs
ArthurHeymans: Done

jay-config/src/input.rs
ArthurHeymans: Done

toml-config/src/lib.rs
ArthurHeymans: Done

mahkoh:
Unfortunately this isn't something that can be implemented reliably at the moment. Layout is almost always performed asynchronously which means that the real position of a node often isn't known at the time the focus changes. I have some code locally that implements layout transactions and which changes layout to be performed synchronously. But it's unclear if I'll continue working on that.

PR #666, opened 2025-11-27
Add cross-output navigation for focus and move commands

ArthurHeymans:
Enable focus-left, focus-right, focus-up, focus-down, move-left, move-right,
move-up, and move-down to cross output boundaries. When reaching the edge of
an output, these commands now find adjacent outputs in the specified direction
and navigate to or move windows to the appropriate edge of the target output's
active workspace.

This is based on top of #665.


mahkoh:
I've implemented a slightly different version in #673. Please check if that works for you.

PR #672, opened 2025-11-28
theme: add bar-position setting

kotarac:
Opted for an enum in case anyone ever decides to add left / right as bar positions.

`Theme` seemed like the most appropriate place to put it in the config as there's already a couple of bar related settings there, while `Status` seemed mostly oriented towards formatting.

Tested against both `WorkspaceDisplayOrder` options to make sure dragging workspaces in `manual` works correctly.

I'm also thinking about adding a setting for hiding the underline (should perhaps rename it now), but probably better to address that separately.

### DIFF
--- a/jay-config/src/theme.rs
+++ b/jay-config/src/theme.rs
@@ -177,6 +177,26 @@ pub fn reset_font() {
     get!().reset_font()
 }
 
+#[non_exhaustive]
+#[derive(Serialize, Deserialize, Debug, Copy, Clone, PartialEq, Eq, Default)]
+pub enum BarPosition {
+    #[default]
+    Top,
+    Bottom,
+}
+
+/// Sets the position of the bar.
+///
+/// Default: `Top`.
+pub fn set_bar_position(position: BarPosition) {
+    get!().set_bar_position(position);
+}
+
+/// Gets the position of the bar.
+pub fn get_bar_position() -> BarPosition {
+    get!(BarPosition::Top).get_bar_position()
+}
+
 /// Elements of the compositor whose color can be changed.
 pub mod colors {
     use {
--- a/src/ifs/wl_surface/tray.rs
+++ b/src/ifs/wl_surface/tray.rs
@@ -135,7 +135,6 @@ impl<T: TrayItem> DynTrayItem for T {
 }
 
 trait TrayItem: Sized + 'static {
-    fn send_initial_configure(&self);
     fn send_current_configure(&self);
     fn data(&self) -> &TrayItemData;
     fn popups(&self) -> &CopyHashMap<XdgPopupId, Rc<Popup<Self>>>;
@@ -353,7 +352,7 @@ fn install<T: TrayItem>(item: &Rc<T>) -> Result<(), TrayItemError> {
     if let Some(node) = data.output.node() {
         data.surface
             .set_output(&node, NodeLocation::Output(node.id));
-        item.send_initial_configure();
+        item.send_current_configure();
     }
     Ok(())
 }
--- a/src/ifs/wl_surface/tray/jay_tray_item_v1.rs
+++ b/src/ifs/wl_surface/tray/jay_tray_item_v1.rs
@@ -9,7 +9,9 @@ use {
                     ack_configure, destroy, get_popup, install,
                 },
             },
-            xdg_positioner::{ANCHOR_BOTTOM_LEFT, ANCHOR_BOTTOM_RIGHT},
+            xdg_positioner::{
+                ANCHOR_BOTTOM_LEFT, ANCHOR_BOTTOM_RIGHT, ANCHOR_TOP_LEFT, ANCHOR_TOP_RIGHT,
+            },
         },
         leaks::Tracker,
         object::{Object, Version},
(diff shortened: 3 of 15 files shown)


mahkoh:
I'll look over this tomorrow but one thing I've seen already is that in jay_tray_item_v1 the preferred anchor and gravity must be adjusted depending on the bar position. I think the current values are also wrong in that LEFT and RIGHT must be swapped.

kotarac:
> I'll look over this tomorrow but one thing I've seen already is that in jay_tray_item_v1 the preferred anchor and gravity must be adjusted depending on the bar position. I think the current values are also wrong in that LEFT and RIGHT must be swapped.

Will take a look. Don't have anything in tray so didn't encounter it.

jay-config/src/theme.rs
```diff
@@ -177,6 +177,25 @@ pub fn reset_font() {
     get!().reset_font()
 }
 
+#[derive(Serialize, Deserialize, Debug, Copy, Clone, PartialEq, Eq, Default)]
+pub enum BarPosition {
+    #[default]
+    Top,
+    Bottom,
+}
```
mahkoh, line 186:
```suggestion
#[non_exhaustive]
#[derive(Serialize, Deserialize, Debug, Copy, Clone, PartialEq, Eq, Default)]
pub enum BarPosition {
    #[default]
    Top,
    Bottom,
}
```

kotarac:
> I've made some changes to master that makes this cleaner.

Hmm, something broke with the bottom bar position after your changes.

Top bar position works ok.

Bottom bar position doesn't show i3status content anywhere. Bar positioning itself and the workspace icons on it are fine though

mahkoh:
This patch should fix it.

```diff
diff --git a/src/renderer.rs b/src/renderer.rs
index 8dec8c2a..5e3c4020 100644
--- a/src/renderer.rs
+++ b/src/renderer.rs
@@ -139,6 +139,8 @@ impl Renderer<'_> {
                         self.state.color_manager.srgb_gamma22(),
                     );
                 }
+                x += bar_rect.x1() - non_exclusive_rect_rel.x1();
+                y += bar_rect.y1() - non_exclusive_rect_rel.y1();
                 if let Some(status) = &rd.status
                     && let Some(texture) = status.tex.texture()
                 {
@@ -159,16 +161,12 @@ impl Renderer<'_> {
                         srgb_srgb,
                     );
                 }
-                {
-                    x += bar_rect.x1() - non_exclusive_rect_rel.x1();
-                    y += bar_rect.y1() - non_exclusive_rect_rel.y1();
-                    for item in output.tray_items.iter() {
-                        let data = item.data();
-                        if data.surface.buffer.is_some() {
-                            let rect = data.rel_pos.get().move_(x, y);
-                            let bounds = self.base.scale_rect(rect);
-                            self.render_surface(&data.surface, rect.x1(), rect.y1(), Some(&bounds));
-                        }
+                for item in output.tray_items.iter() {
+                    let data = item.data();
+                    if data.surface.buffer.is_some() {
+                        let rect = data.rel_pos.get().move_(x, y);
+                        let bounds = self.base.scale_rect(rect);
+                        self.render_surface(&data.surface, rect.x1(), rect.y1(), Some(&bounds));
                     }
                 }
             }
```

kotarac:
> This patch should fix it.

Thanks, testing that and the tray fixes a little bit and then pushing for review.

src/ifs/wl_surface/tray/jay_tray_item_v1.rs
```diff
@@ -107,12 +122,12 @@ impl JayTrayItemV1RequestHandler for JayTrayItemV1 {
 
 impl TrayItem for JayTrayItemV1 {
     fn send_initial_configure(&self) {
-        self.send_preferred_anchor();
-        self.send_preferred_gravity();
         <Self as TrayItem>::send_current_configure(self);
     }
```
mahkoh: This function can be removed.

src/ifs/wl_surface/tray/jay_tray_item_v1.rs
```diff
@@ -107,12 +122,12 @@ impl JayTrayItemV1RequestHandler for JayTrayItemV1 {
...
 
     fn send_current_configure(&self) {
```
mahkoh, line 120: Pass the output to this function as an argument.

src/ifs/wl_surface/tray.rs
```diff
@@ -135,8 +135,7 @@ impl<T: TrayItem> DynTrayItem for T {
 }
 
 trait TrayItem: Sized + 'static {
-    fn send_initial_configure(&self);
-    fn send_current_configure(&self);
+    fn send_current_configure(&self, output: &Rc<OutputGlobalOpt>);
```
mahkoh: All of these function should take the `OutputNode` directly as an argument. No need to look at `self.data`.

kotarac: How about just decoupling all of it from output like it was before?

mahkoh: You're right. I just saw that you were using the output, got confused, and thought that the bar position was a per-output setting. The current version is fine.

PR #673, opened 2025-11-29
tree: allow moving focus between containers on different outputs

mahkoh:
Closes #666
Closes #665


ArthurHeymans:
> Floating windows are always tiled on the output that contains the cursor.

Oh you're right! This looks good then.

PR #686, opened 2025-12-14
vulkan: don't swap order of texture operations

### LINKED ISSUE #685: Random graphical errors after a long time of use
Graphical errors include random window darkening and random black boxes. The darkening is exclusive to windows and seems to go away when the window is focused, the random black boxes affect both windows and random locations on the screen. Usually switching to the workspaces, focusing the window, or moving my mouse to that area makes it go away, so does switching to OpenGL.

As far as I'm aware, this only happens under Vulkan but this could happen under OpenGL as well and I just don't know about it as I pretty much only use Jay with Vulkan. I'm using a 6700xt and `vulkan-radeon`.

I'm not sure what else I could add to help reproduce this issue for debugging so let me know. I'll also add screenshots / recordings after I manage to reproduce this again. It tends to start happening after several hours of use.

mahkoh:
Fixes #685

### DIFF
--- a/src/gfx_apis/vulkan/renderer.rs
+++ b/src/gfx_apis/vulkan/renderer.rs
@@ -201,6 +201,7 @@ enum VulkanOp {
 
 struct VulkanTexOp {
     tex: Rc<VulkanImage>,
+    index: usize,
     range: Range<usize>,
     buffer_resv: Option<Rc<dyn BufferResv>>,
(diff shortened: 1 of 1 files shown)


Stoppedpuma:
@mahkoh I haven't seen any of the graphical errors with this PR's changes after usage for about 4-5ish hours. I can test for longer as well if you would like, otherwise this seems to solve the issue.

Issue #678, opened 2025-12-02
more granular show-titles setting


kotarac:
With `show-titles = false`, it's hard to visually distinguish a focused window.

Sway partially mitigates this by having a different border color for focused windows. I have a naive implementation for this, but it seems pretty ugly even with not much code and it's not the nicest thing anyway, so it wasn't something I wanted to pursue.

Another issue is it equally hides titles for both normal and floating windows.
Sway has `default_border` and `default_floating_border` settings to mitigate this.

I wouldn't necessarily want to emulate Sway in this regard, so I was considering alternatives.

A potential solution could be making `show-titles` have multiple options:
- `always` - always shows titles, matching behavior of current `true`
- `multiple` - only shows titles if there's more than 1 window on the output
- `floating-and-multiple` - shows titles on floating windows and/or when there's more than 1 window on the output
- `floating` - shows titles only on floating windows
- `never` - never shows titles, matching behavior of current `false`

However, the stated goal of Jay is:
> "Jay also aims to be forward and backward compatible for existing setups, allowing you to upgrade or downgrade the compositor without having to adjust your configuration."

I consider that very valuable, so I'm not sure how to go about this.
Adding another option (implementing these options if `show-titles = true`, while keeping `show-titles` untouched seems a bit messy.

No strong opinions, would be happy to hear thoughts.

kotarac:
> I would like to understand why people are using `show-titles = false` in the first place.

Sure. In my case it's pretty simple and stems from using bar-position bottom like I'm used to in i3 and Sway.

If you have something on an output with controls on top, if there's nothing above it's very hard to missclick anything. If titles are enabled then you're aiming between controls and the title (or bar with bar-position top).
An obvious example are browsers like Chrome and Firefox with tabs on top. Both have a clickable area in tabs all the way to the top, which in this case ends up being on the very top of output, making it hard to miss.

Obviously, with a very keyboard focused workflow, this isn't a very big deal, but it makes a difference.

Other than that, it's also a matter of visual taste. I prefer no title bar when there's nothing else on the output because it doesn't really serve anything.


Regarding the config proposal, since then I've done a naive implementation that I want to refine a little bit before getting it to you (there are some caveats with multiple and counting both tiled and floating windows).
However, I've been considering a simpler approach of just having `true`, `smart`, `false` to avoid having config overload. `smart` essentially being `floating-or-multiple`.

kotarac:
> How about moving titles to the bottom as well?

[...]

PR #702, opened 2025-12-20
config: add show-titles auto option

kotarac:
Covers #678, but simpler than the original proposal there. Opted to change `show-titles` instead of going with window rules and new simple actions.

It just adds a new `auto` option that shows titles if there's more than 1 window in a container or if the window is floating. Intentionally didn't want to complicate it by counting windows on workspace (in a container + any floating) because that seemed too complicated for no value.

~Toggle action rotates between `show-titles` options. Wasn't sure about it, I guess it could also skip `auto`.~

Also fixes an obscure crash when borders are larger than the window after `border-width` gets changed dynamically (changing config and reloading) and covers that with an integration test.


jay-config/src/_private.rs
```diff
@@ -15,7 +15,7 @@ use {
     std::marker::PhantomData,
 };
 
-pub const VERSION: u32 = 1;
+pub const VERSION: u32 = 2;
```
mahkoh: This would break every config.so out there. Instead you have to add new messages and do feature detection as in https://github.com/mahkoh/jay/blob/867172f20bbf3bdeee1af1a64314d5cf5e390a6c/jay-config/src/_private/client.rs#L1521-L1533

jay-config/src/lib.rs
```diff
@@ -346,19 +347,27 @@ pub fn toggle_show_bar() {
...
-pub fn set_show_titles(show: bool) {
-    get!().set_show_titles(show)
+pub fn set_show_titles(show: impl Into<ShowTitles>) {
+    get!().set_show_titles(show.into())
 }
 
 /// Returns whether title bars on windows are shown.
-pub fn get_show_titles() -> bool {
-    get!(true).get_show_titles()
+pub fn get_show_titles() -> ShowTitles {
```
mahkoh: You also can't change the signatures of existing functions. Instead you have to add new functions.

jay-config/src/lib.rs
```diff
@@ -346,19 +347,27 @@ pub fn toggle_show_bar() {
...
+    get!(ShowTitles::True).get_show_titles()
 }
 
 /// Toggles whether title bars on windows are shown.
+///
+/// Rotates between `True`, `False`, and `Auto`.
 pub fn toggle_show_titles() {
```
mahkoh: This function needs to continue toggling between true and false.

jay-config/src/_private.rs
kotarac: Will do.

Do you think it would make sense to have an example config shared lib and an integration test to cover this? I keep forgetting about this feature, to be honest. :slightly_smiling_face:

jay-config/src/lib.rs
kotarac: Because this would break if someone was using get_show_titles in a config shared library or is there an another concern?

mahkoh: You're correct.

jay-config/src/lib.rs
```diff
@@ -346,19 +347,44 @@ pub fn toggle_show_bar() {
 /// Sets whether title bars on windows are shown.
 ///
 /// The default is `true`.
+///
+/// See also [`set_show_titles_v2`].
+#[deprecated(note = "use `set_show_titles_v2` instead")]
```
mahkoh, line 352: I don't think there is a need to mark this as deprecated.

jay-config/src/lib.rs
```diff
@@ -346,19 +347,44 @@ pub fn toggle_show_bar() {
...
+/// See also [`get_show_titles_v2`].
+#[deprecated(note = "use `get_show_titles_v2` instead")]
 pub fn get_show_titles() -> bool {
     get!(true).get_show_titles()
 }
 
+/// Sets when title bars on windows are shown.
+///
+/// The default is `True`.
+pub fn set_show_titles_v2(show: impl Into<ShowTitles>) {
```
mahkoh, line 368: Not a fan of _v2.

PR #705, opened 2025-12-26
config: Add keymap_from_names for RMLVO keymaps

khyperia:
I find RMLVO keymaps much more approachable than the XKB text file format.

libxkbcommon allows you to create a keymap with RMLVO names with [`xkb_keymap_new_from_names2`](https://xkbcommon.org/doc/current/group__keymap.html#gab500a4d834bb839067c5c5fc2144d457), and handily, the Rust libxkbcommon reimplementation we use also supports them, with [`Context::keymap_from_names`](https://docs.rs/kbvm/latest/kbvm/xkb/struct.Context.html#method.keymap_from_names). So, we just have to pipe the arguments through from `jay-config` to `kbvm`.

Strictly, this PR is not necessary for users to use RMLVO to define their keymaps: they could reference the kbvm crate themselves, create a keymap with `keymap_from_names`, format it to a string with `format!("{}", keymap.format())`, and pass that string to Jay's `parse_keymap`, but that's a bit silly IMO!

Implementing this in `jay-toml-config` could be done in a future PR if there's a user who wants it.

### DIFF
--- a/src/cli/input.rs
+++ b/src/cli/input.rs
@@ -77,6 +77,8 @@ pub enum SeatCommand {
     SetRepeatRate(SetRepeatRateArgs),
     /// Set the keymap.
     SetKeymap(SetKeymapArgs),
+    /// Set the keymap from RMLVO names.
+    SetKeymapFromNames(SetKeymapFromNamesArgs),
     /// Retrieve the keymap.
     Keymap,
     /// Configure whether this seat uses the hardware cursor.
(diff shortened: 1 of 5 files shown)


mahkoh:
This looks good overall. I'll look into it in more detail when I'm back home. Just one thing: can't you calculate the blake3 hash in one go on the rmgo tuple?

khyperia:
take your time! no rush.

> can't you calculate the blake3 hash in one go on the rmgo tuple?

There might be something I'm missing, but from my understanding, the blake3 crate only takes `&[u8]`, not arbitrary data - it doesn't implement `std::hash::Hasher`, so you can't throw a tuple of `std::hash::Hash` objects at it. I may be missing something though, it's been a while since I've written Rust code!

mahkoh:
Ah you're right.

PR #709, opened 2025-12-29
rect: safer construction

kotarac:
Fixes the case I mentioned in another PR, wanted to do a proper fix and have it separate.

Easiest way to reproduce is to resize the window as small as it goes and then change `border-width` or `title-height` to something big in runtime (e.g. by reloading TOML config).
It's a bit of an edge case (pun intended), but a crash is a crash. :slightly_smiling_face: 

EDIT: updated PR title and the approach with 2 new functions after the discussion in comments.


kotarac:
> x1 <= x2 and y1 <= y2 are invariants of Rect. Relying on that existing validation seems like the right choice to me.

I've been reading up a bit more. Let me see if I understand.

We want to use the new functions (2 unchecked variants that just always return `Rect`) for "internal" / "trusted" cases, where input is 100% controlled (through config limitations, etc), or when we simply know clamping is fine and we don't care how it got to it.
In other words, as you said, where we were unwrapping because it should be correct (except for bugs like the one fixed in this PR).

We want to use the existing checked functions returning `Option<Rect>` in other places because we want to be explicit about handling any errors in the "untrusted" cases like when handling invalid input.

TL;DR separating logic errors from input errors.

mahkoh:
I made a small change where I restored the `.max(0)` when calculating the output rects. I did not want to validate the changed behavior.

PR #713, opened 2026-01-05
implement wlr-gamma-control-unstable-v1

### LINKED ISSUE #328: wlr-gamma-control protocol
Are there any plans to support wlr-gamma-control?

khyperia:
A draft PR because:

- This only implements it on hardware with the `GAMMA_LUT` property, there is no software rendering fallback
  - ~~The zwlr_gamma_control_manager_v1 global is not exposed if the hardware doesn't support it~~ (update: it's always exposed, but emits the failed event if hardware doesn't support it)
- As mentioned in https://github.com/mahkoh/jay/issues/328, `GAMMA_LUT` is soon to be deprecated, replaced by `COLOR_PIPELINE` (I believe in a few months, with Linux 7.0, seems like?)

Regardless, my sleep is being wrecked by my bright blue screen and I want `wlsunset`/`gammastep`, haha, so I wrote this up for myself and anyone else who might want to run a patched Jay until `COLOR_PIPELINE` releases. If you think this PR is good, though, feel free to merge it! It's at least a nice base to work off of when implementing `COLOR_PIPELINE`.

Closes #328

### DIFF
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -37,7 +37,7 @@ use {
         hash::Hash,
         rc::Rc,
     },
-    uapi::{OwnedFd, c},
+    uapi::{OwnedFd, Packed, Pod, c},
 };
 
 pub mod transaction;
@@ -164,6 +164,9 @@ pub trait Connector: Any {
         Err(BackendConnectorTransactionError::TransactionsNotSupported(
             self.kernel_id(),
         ))
+    }
+    fn gamma_lut_size(&self) -> Option<u32> {
+        None
     }
 }
 
@@ -611,6 +614,42 @@ impl BackendColorSpace {
     }
 }
 
+// kernel: struct drm_color_lut
+#[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Hash)]
+#[repr(C)]
+pub struct BackendGammaLutElement {
+    pub red: u16,
+    pub green: u16,
+    pub blue: u16,
+    pub reserved: u16,
+}
+
+unsafe impl Pod for BackendGammaLutElement {}
+unsafe impl Packed for BackendGammaLutElement {}
+
+#[derive(Debug, Eq)]
+pub struct BackendGammaLut {
+    id: [u8; 32],
+    pub gamma_lut: Vec<BackendGammaLutElement>,
+}
+
+impl BackendGammaLut {
+    pub fn new(mut gamma_lut: Vec<BackendGammaLutElement>) -> Self {
+        for element in &mut gamma_lut {
+            element.reserved = 0;
+        }
+        let gamma_lut_bytes = uapi::as_bytes(&gamma_lut as &[_]);
+        let id = *blake3::hash(gamma_lut_bytes).as_bytes();
+        Self { id, gamma_lut }
+    }
+}
+
+impl PartialEq for BackendGammaLut {
+    fn eq(&self, other: &Self) -> bool {
+        self.id == other.id
+    }
+}
+
 linear_ids!(
     BackendConnectorStateSerials,
     BackendConnectorStateSerial,
@@ -629,4 +668,5 @@ pub struct BackendConnectorState {
     pub format: &'static Format,
     pub color_space: BackendColorSpace,
     pub eotf: BackendEotfs,
+    pub gamma_lut: Option<Rc<BackendGammaLut>>,
 }
--- a/src/backends/metal/transaction.rs
+++ b/src/backends/metal/transaction.rs
@@ -2,7 +2,8 @@ use {
     crate::{
         allocator::BufferObject,
         backend::{
-            BackendColorSpace, BackendConnectorState, BackendEotfs, Connector, ConnectorEvent,
+            BackendColorSpace, BackendConnectorState, BackendEotfs, BackendGammaLut, Connector,
+            ConnectorEvent,
             transaction::{
                 BackendAppliedConnectorTransaction, BackendConnectorTransaction,
                 BackendConnectorTransactionError, BackendPreparedConnectorTransaction,
@@ -57,6 +58,9 @@ pub struct DrmCrtcState {
     pub mode_blob: Option<Rc<PropBlob>>,
     pub vrr_enabled: bool,
     pub assigned_connector: DrmConnector,
+    pub gamma_lut: Option<Rc<BackendGammaLut>>,
+    pub gamma_lut_blob_id: DrmBlob,
+    pub gamma_lut_blob: Option<Rc<PropBlob>>,
 }
 
 #[derive(Default, Clone, Debug)]
@@ -420,6 +424,22 @@ impl MetalDeviceTransaction {
                 crtc.new.mode_blob = Some(Rc::new(blob));
                 mode.clone()
             };
+            if crtc.new.gamma_lut != state.gamma_lut {
+                if let Some(gamma_lut) = &state.gamma_lut {
+                    let blob = slf
+                        .dev
+                        .dev
+                        .master
+                        .create_blob(&gamma_lut.gamma_lut as &[_])
+                        .map_err(BackendConnectorTransactionError::CreateGammaLutBlob)?;
+                    crtc.new.gamma_lut_blob_id = blob.id();
+                    crtc.new.gamma_lut_blob = Some(Rc::new(blob));
+                } else {
+                    crtc.new.gamma_lut_blob_id = DrmBlob::NONE;
+                    crtc.new.gamma_lut_blob = None;
+                }
+                crtc.new.gamma_lut = state.gamma_lut.clone();
+            }
             for plane_id in [&mut crtc_planes.primary, &mut crtc_planes.cursor] {
                 if plane_id.is_none() {
                     continue;
@@ -840,6 +860,12 @@ impl MetalDeviceTransactionWithDrmState {
                 if n.mode_blob_id != o.mode_blob_id {
                     log_change!(o, n, mode_blob_id);
                     c.change(crtc.obj.mode_id, n.mode_blob_id);
+                }
+                if let Some(gamma_lut) = crtc.obj.gamma_lut
+                    && n.gamma_lut_blob_id != o.gamma_lut_blob_id
+                {
+                    log_change!(o, n, gamma_lut_blob_id);
+                    c.change(gamma_lut, n.gamma_lut_blob_id);
                 }
                 reset_default_properties!(
                     c,
--- a/src/clientmem.rs
+++ b/src/clientmem.rs
@@ -16,7 +16,7 @@ use {
     },
     thiserror::Error,
     uapi::{
-        OwnedFd,
+        OwnedFd, Pod,
         c::{self, raise},
         ftruncate,
     },
@@ -30,6 +30,8 @@ pub enum ClientMemError {
     Sigbus,
     #[error("mmap failed")]
     MmapFailed(#[source] crate::utils::oserror::OsError),
+    #[error("Length was not a multiple of the data element size")]
+    InvalidLength,
 }
 
 pub struct ClientMem {
@@ -182,13 +184,17 @@ impl ClientMemOffset {
         }
     }
 
-    pub fn read(&self, dst: &mut Vec<u8>) -> Result<(), ClientMemError> {
+    pub fn read<T: Pod>(&self, dst: &mut Vec<T>) -> Result<(), ClientMemError> {
+        if self.data.len().checked_rem(std::mem::size_of::<T>()) != Some(0) {
+            return Err(ClientMemError::InvalidLength);
+        }
         self.access(|v| {
-            dst.reserve(v.len());
-            let (_, unused) = dst.split_at_spare_mut_ext();
+            let len_elements = v.len() / std::mem::size_of::<T>();
+            dst.reserve(len_elements);
+            let (_, unused) = dst.split_at_spare_mut_bytes_ext();
             unused[..v.len()].copy_from_slice(uapi::as_maybe_uninit_bytes(v));
             unsafe {
-                dst.set_len(dst.len() + v.len());
+                dst.set_len(dst.len() + len_elements);
             }
         })
     }
--- a/src/ifs/zwlr_gamma_control_manager_v1.rs
+++ b/src/ifs/zwlr_gamma_control_manager_v1.rs
@@ -1,0 +1,115 @@
+use {
+    crate::{
+        client::{CAP_GAMMA_CONTROL_MANAGER, Client, ClientCaps, ClientError},
+        globals::{Global, GlobalName},
+        ifs::zwlr_gamma_control_v1::*,
+        leaks::Tracker,
+        object::{Object, Version},
+        wire::{ZwlrGammaControlManagerV1Id, zwlr_gamma_control_manager_v1::*},
+    },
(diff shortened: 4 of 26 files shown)


khyperia:
Some implementation notes:

There's a handful of ways I could foresee the LUT being plumbed through to the backend. I went for putting an `Rc<Option<Vec<_>>>` into `BackendConnectorState`, to pass the LUT "by value" along with all the other backend state. `Rc` for efficiency, and using `Rc::ptr_eq` to do a quick comparison to see whether a new LUT blob should be created. One downside of this approach is that `BackendConnectorState` is no longer `Copy`, it is merely `Clone`, which required quite a few scattered changes to add `.clone()`s instead of copies and whatnot. In the future, once `COLOR_PIPELINE` releases, I imagine this will turn into `Rc<ColorPipeline>`, where `ColorPipeline` is, say, an enum that could be a CTM, a LUT, a combination/sequence of ColorPipelines, etc.

Another option is to add an additional method(s) called `.set_gamma_lut(...)` on the backend (or something like that), *not* passing through `BackendConnectorState`. The advantage of this is that the "heavy" LUT data does not have to go through the by-value passing of `BackendConnectorState`, and allows it to remain `Copy`. This doesn't take advantage of the existing atomic commit infrastructure though, requiring some finangling.

src/ifs/zwlr_gamma_control_v1.rs
```diff
@@ -0,0 +1,166 @@
...
+
+        let Some(node) = self.output.node() else {
+            slf.send_failed();
+            return Ok(());
+        };
+
+        if let Some(old) = node.active_zwlr_gamma_control.get() {
+            if self.id() != old {
+                slf.send_failed();
+                return Ok(());
```
khyperia: Calling out a small design choice I made here: What happens when two clients both try to set gamma for the same output?

- you could fail the second one, keeping the first one with exclusive control
- you could fail the first one, replacing it whenever a new request comes in
- you could combine the two?? or something??

Compositors can do whatever they feel like, according to spec. Wlroots appears to choose the first option, so that's what I did too

mahkoh: >There can only be at most one gamma control object per output, which has exclusive access to this particular output.

zwlr_gamma_control_manager_v1.get_gamma_control must check that no other object exists.

This whole block is unnecessary if that request already reserves exclusive access to the output.

src/ifs/zwlr_gamma_control_v1.rs
```diff
@@ -0,0 +1,166 @@
...
+    thiserror::Error,
+};
+
+pub struct ZwlrGammaControlV1 {
+    id: ZwlrGammaControlV1Id,
+    client: Rc<Client>,
+    version: Version,
+    output: Rc<OutputGlobalOpt>,
+    pub tracker: Tracker<Self>,
+}
```
mahkoh, line 23: This object must contain a flag that says if the object has ownership of the output. Sending the failed event should set the flag to false.

src/ifs/zwlr_gamma_control_v1.rs
```diff
@@ -0,0 +1,166 @@
...
+                node.set_gamma_lut(Rc::new(None));
+            }
+        };
+    }
+
+    // Wayland's LUT is ([red], [green], [blue]). DRM's LUT is [(red, green, blue, _)]. Both are u16.
+    fn wayland_gamma_lut_to_drm_gamma_lut(data_u8: &[u8]) -> Vec<BackendGammaLutElement> {
+        let data = unsafe {
+            std::slice::from_raw_parts(data_u8.as_ptr() as *const u16, data_u8.len() / 2)
+        };
```
mahkoh: This is unsound because the pointer isn't necessarily aligned to 16 bits. Instead, change ClientMemOffset::read to allow reading into an arbitrary `Vec<T: Pod>` with a check that the buffer size is a multiple of sizeof(T).

src/ifs/zwlr_gamma_control_manager_v1.rs
```diff
@@ -0,0 +1,111 @@
...
+        1
+    }
+
+    fn exposed(&self, state: &State) -> bool {
+        state
+            .outputs
+            .lock()
+            .values()
+            .any(|output| output.connector.connector.gamma_lut_size().is_some())
+    }
```
mahkoh: I'd prefer to just remove this. This is handled by the failed event.

src/utils/clonecell.rs
```diff
@@ -104,3 +105,5 @@ unsafe impl UnsafeCellCloneSafe for Modifiers {}
 unsafe impl UnsafeCellCloneSafe for NodeId {}
 
 unsafe impl UnsafeCellCloneSafe for Window {}
+
+unsafe impl UnsafeCellCloneSafe for BackendConnectorState {}
```
mahkoh: I'd prefer not to do this. Let's just use RefCell where necessary.

src/video/drm/sys.rs
```diff
@@ -917,9 +917,9 @@ struct drm_mode_create_blob {
 
 const DRM_IOCTL_MODE_CREATEPROPBLOB: u64 = drm_iowr::<drm_mode_create_blob>(0xbd);
 
-pub fn mode_create_blob<T>(fd: c::c_int, t: &T) -> Result<DrmBlob, OsError> {
+pub fn mode_create_blob<T: Pod + ?Sized>(fd: c::c_int, t: &T) -> Result<DrmBlob, OsError> {
```
mahkoh:
```suggestion
pub fn mode_create_blob<T: ?Sized>(fd: c::c_int, t: &T) -> Result<DrmBlob, OsError> {
```

src/backend.rs
```diff
@@ -612,13 +615,24 @@ impl BackendColorSpace {
     }
 }
 
+// kernel: struct drm_color_lut
+#[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Hash)]
+pub struct BackendGammaLutElement {
```
mahkoh, line 620:
```suggestion
#[repr(C)]
pub struct BackendGammaLutElement {
```

src/ifs/zwlr_gamma_control_v1.rs
khyperia: Aah, right, `get_gamma_control` should return an object that immediately has a failed event, rather than emitting a failed event upon attempting to `set_gamma`.

Tbh I kind of disagree with the spec here, if a compositor decides to allow a new `zwlr_gamma_control_v1` to replace an old one, the old one shouldn't be failed until the new one has provided a `set_gamma` - otherwise, there's a potential of a flash of no gamma being applied, if the new one doesn't call `set_gamma` fast enough. But yeah, I'll implement the spec as written!

src/video/drm/sys.rs
khyperia: Are you sure? I added this after realizing that I was passing in a `&Vec<_>` to this function, which is very different (and spooky) from what I wanted, which was passing in a `&[_]` (it would create a blob containing the ptr+capacity+length data, rather than the data itself)

src/ifs/zwlr_gamma_control_v1.rs
khyperia: There's a bit of spec wonkiness here:

> At any time the compositor can send a failed event indicating that this object is no longer valid.
> [The failed event] can happen for a number of reasons, including: Setting the gamma tables failed

Ok, cool, if the client sends bad `set_gamma` data, it sends a failed event.

> When the gamma control object is destroyed, the gamma table is restored to its original value.

This implies that if the client previously sent good `set_gamma` data, that data should be kept, until the object is destroyed

> Destroy event: If the object is still valid, this restores the original gamma tables.

But, wait, uuuh, huh?? "If the object is still valid", meaning we *don't* restore if the object has failed?

I believe the spec should be changed to

> When the gamma control object is destroyed [or is no longer valid], the gamma table is restored to its original value.

and that's what I'll implement

src/ifs/zwlr_gamma_control_v1.rs
khyperia: When looking into this, I found that `req.fd.borrow().read_exact` seems much simpler than going through ClientMem - using read() instead of mmap -> manual copy. To me, ClientMem machinery makes sense for high performance repeated reads, etc. (e.g. WlShm buffers), but just reading from the fd for basic stuff is more straightforward. Let me know if you disagree and I can change it back!

(Adding Pod support, rather than just u8, for ClientMem, is a bit hairy for me - lots of unsafe going around, and some relevant uapi and/or MaybeUninit helpers missing)

mahkoh: Unfortunately using read_exact is not correct because file description offset is not guaranteed to be at the start of the file or to remain stable. You'd have to use the equivalent of `pread(2)` to make this work. I think using ClientMemOffset is simpler.

src/video/drm/sys.rs
mahkoh: `T: Pod` means that transmuting arbitrary bytes to `T` produces a valid object. Since we're not creating a new `T` here, the bound is not appropriate.

The only other trait in uapi for this purpose is `T: Packed` which indicates that `T` has no padding bytes and therefore transmuting `T` to bytes produces valid bytes. However, the blob API can deal with invalid bytes so it's not appropriate here either.

If you want to add some safety to this, you can create a new, safe, marker trait `DrmBlobData` that types that are supposed to be passed to this function can implement.

src/ifs/zwlr_gamma_control_manager_v1.rs
```diff
@@ -0,0 +1,112 @@
...
+
+impl Global for ZwlrGammaControlManagerV1Global {
+    fn singleton(&self) -> bool {
+        true
+    }
+
+    fn version(&self) -> u32 {
+        1
+    }
+}
```
mahkoh, line 61:
```suggestion
impl Global for ZwlrGammaControlManagerV1Global {
    fn singleton(&self) -> bool {
        true
    }

    fn version(&self) -> u32 {
        1
    }
    
    fn required_caps(&self) -> ClientCaps {
        // ...
    }
}
```

src/backends/metal/transaction.rs
```diff
@@ -55,6 +56,9 @@ pub struct DrmCrtcState {
     pub mode_blob: Option<Rc<PropBlob>>,
     pub vrr_enabled: bool,
     pub assigned_connector: DrmConnector,
+    pub gamma_lut: Rc<Option<Vec<BackendGammaLutElement>>>,
```
mahkoh:
```suggestion
    pub gamma_lut: Option<Rc<BackendGammaLut>>,
```

Where BackendGammaLut has a cheap PartialEq implementation. Let's pre compute blake3 on the elements.

src/clientmem.rs
```diff
@@ -182,13 +184,17 @@ impl ClientMemOffset {
         }
     }
 
-    pub fn read(&self, dst: &mut Vec<u8>) -> Result<(), ClientMemError> {
+    pub fn read<T: Pod>(&self, dst: &mut Vec<T>) -> Result<(), ClientMemError> {
```
mahkoh, line 187:
```suggestion
    pub fn read<T: Pod>(&self, dst: &mut Vec<T>) -> Result<(), ClientMemError> {
        if size_of::<T>() == 0 { return error; }
```

src/tree/output.rs
```diff
@@ -125,6 +125,7 @@ pub struct OutputNode {
     pub ext_workspace_groups: CopyHashMap<WorkspaceManagerId, Rc<ExtWorkspaceGroupHandleV1>>,
     pub pinned: LinkedList<Rc<dyn PinnedNode>>,
     pub tearing: Cell<bool>,
+    pub has_valid_zwlr_gamma_control: Cell<bool>,
```
mahkoh: This should instead be a reference to the object itself. You need to send failed when the output is disconnected.

src/ifs/zwlr_gamma_control_manager_v1.rs
khyperia: I looked into the ClientCaps system a bit and I'm not quite sure I understand it enough to implement this. Do I add a new ClientCap? What even are ClientCaps? (AFAIK it's a Jay-specific thing, perhaps related to the privileged system?)

mahkoh: Capabilities restrict which clients can access which globals. Clients that connect to `wayland-N.jay` automatically have all capabilities. Otherwise capabilities can be added via client rules.

You just have to add a new enum variant and then also add the new variant in the config and toml config.

src/clientmem.rs
khyperia: This is already checked via the `checked_rem` in the line below (`if self.data.len().checked_rem(std::mem::size_of::<T>()) != Some(0)`) - `checked_rem` returns `None` if the denominator is zero.

src/backends/metal/transaction.rs
khyperia: Hmm... I'll take this as an opportunity to refactor a bit to refactor a bit to prepare for the upcoming GAMMA_LUT -> COLOR_PIPELINE drm property deprecation/change: instead of an `Option<Rc<BackendGammaLut>>`, it could be a `Rc<ColorPipeline>`. In the future, a ColorPipeline struct could contain an (ordered) list of (optional) transformations:

- A gamma LUT
- An inverse gamma LUT
- A color transformation matrix
- ... and more...

Anything that the COLOR_PIPELINE property can contain.

In the present, the ColorPipeline struct only contains an optional gamma LUT, because it currently maps onto the GAMMA_LUT property, rather than COLOR_PIPELINE.

Feel free to tell me I'm putting the cart before the horse, and that I should keep things simple with a `Option<Rc<BackendGammaLut>>`, until we have a use case and implementation for the full COLOR_PIPELINE property!

src/backend.rs
```diff
@@ -612,13 +615,46 @@ impl BackendColorSpace {
...
+    pub reserved: u16,
+}
+
+unsafe impl Pod for BackendGammaLutElement {}
+unsafe impl Packed for BackendGammaLutElement {}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash, Default)]
+pub struct ColorPipelineId([u8; 32]);
+
+#[derive(Debug, Eq, PartialEq, Default)]
```
mahkoh: PartialEq needs to be implemented manually to only compare the ID.

src/backends/metal/transaction.rs
mahkoh: I don't believe that is how it's going to work. The pipeline will be programmed transparently by the metal backend as an optimization when possible. It's not something that will be controlled explicitly by the rest of the compositor. Similar to how direct scanout is a transparent optimization. For example, if direct scanout would otherwise be impossible because the buffer is encoded as sRGB while the monitor is using PQ, the backend might program the pipeline to do this conversion in hardware instead of performing the conversion in a shader.

I think this:

```rust
pub struct BackendColorPipeline {
    id: ColorPipelineId,
    pub gamma_lut: Option<Vec<BackendGammaLutElement>>,
}
```

should just be

```rust
pub struct BackendGammaLut {
    hash: [u8; 32],
    pub gamma_lut: Vec<BackendGammaLutElement>,
}
```

for now.

khyperia:
Your changes work great! Fantastic, thank you :sparkles:

Issue #718, opened 2026-01-17
vrr cursor-hz option to match content fps


kotarac:
Opening this to discuss if this makes sense, if it could be implemented at all and what would be the best way to go about it.

Currently, the `cursor-hz` option "Limits the rate at which cursors are updated on screen when VRR is active".

Ideally, IMO, there would be an option that makes it match the FPS of the content being rendered, down to some minimum (either configurable, or based on the screen's VRR range, maybe taking LFC into account somehow).

I have seen this issue being mentioned on almost every wayland compositor I know of in various places, but I've only personally seen it in Sway (https://github.com/swaywm/sway/issues/6832).
Considering Jay has `cursor-hz`, I assume you've seen this mentioned as well.

On Windows, IIRC, this works as I described in the ideal case above. Admittedly, the last time I used Windows was before 2020, so my memory could be wrong, but Google seems to agree.

Do you think this would be feasible in Jay? On a high level, what would be the ideal way to go about it?

kotarac:
> Doesn't setting the cursor hz to 1 already do this?

Wouldn't 1 equal a maximum refresh rate from mouse movement of 1 hz?

Admittedly, I haven't tested the value of `1` because that seemed weird. 🙂

kotarac:
For some reason, I can't get VRR to work in a game now to test it properly (it's just on max refresh rate all the time despite FPS being lower).

But I did observe the mouse being smooth with a game being focused, so I guess that does work.

So if I'm understanding this correctly, refresh rate with `cursor-hz` on and VRR enabled is essentially `max(min(cursor-hz, current cursor polling rate), current content FPS)`, ignoring other hardware constraints like VRR range etc?

And the diff in behavior from Windows that I'm seeing is that Win probably does that only while the game is focused or something along those lines.

FWIW, in my testing, VRR variants 1 and 2 applied to the whole output if there's fullscreen windows anywhere, not just when they're focused.

mahkoh:
>So if I'm understanding this correctly, refresh rate with cursor-hz on and VRR enabled is essentially max(min(cursor-hz, current cursor polling rate), current content FPS), ignoring other hardware constraints like VRR range etc?

The cursor is updated whenever the rest of the screen is updated but no later than 1/cursor-hz after the last screen update.

>And the diff in behavior from Windows that I'm seeing is that Win probably does that only while the game is focused or something along those lines.

cursor-hz is used whenever VRR is active on the screen.

kotarac:
[...]

PR #725, opened 2026-01-19
docs: vrr clarifications

kotarac:
Tried hard to figure out what led me to series of failed attempts and wrong conclusions (other than the bug) in #718 and updated the docs in the hopes it will help others setting up VRR in Jay.

Let me know if the wording works and if it's correct / precise enough while still being valuable to non-technical folks.

Some thoughts below.

For `VrrMode::VARIANT_3` I wanted it to be explicit it's up to the application to report the content type. My 1st thought when seeing this option was "how would Jay know?" and I needed to check the implementation to understand there's no magical detection here.
Noting the exact wayland protocol might be too much though, not sure.

For `cursor-hz` I wanted multiple things clarified.
When we discussed it in the issue, at 1st I thought the value of `1` was a special case until I went to check the implementation.
The previous wording in a couple of places implied (at least IMO) that cursor updates would be limited to that refresh rate, while in fact cursor will update more often than `cursor-hz` if the content refresh rate is higher than `cursor-hz`.
There's also the implied, but not obvious, relationship of mouse polling rate, `cursor-hz`, content refresh rate that I feel like deserved a clarification.
Lastly, for majority of users, setting it to some reasonable value (like your example of `cursor-hz = 90`) probably makes most sense, so I thought it deserved to be explained why.

Not sure about the term `content refresh rate`.

References to `mouse` are perhaps not ideal, because technically the same would apply to any pointer device.

Also considered explaining why `VrrMode::ALWAYS` might not make sense, but wasn't sure how to phrase that. To me at least it was clear I didn't want that, other than for testing, but I imagine it's not obvious. Although, `always` with a high enough `cursor-hz` might even be okish for terminal/browser usage.

Love how well it works since I set it up, thank you for building this!

### DIFF
--- a/toml-spec/spec/spec.generated.json
+++ b/toml-spec/spec/spec.generated.json
@@ -2094,22 +2094,22 @@
           "$ref": "#/$defs/VrrMode"
         },
         "cursor-hz": {
-          "description": "The VRR cursor refresh rate.\n\nLimits the rate at which cursors are updated on screen when VRR is active.\n",
+          "description": "The VRR cursor refresh rate.\n\nLimits the rate at which cursor movement forces a screen update\nwhen VRR is active.\n",
           "$ref": "#/$defs/VrrHz"
         }
       },
       "required": []
     },
     "VrrHz": {
-      "description": "A VRR refresh rate limiter.\n\n- Example 1:\n\n  ```toml\n  vrr = { cursor-hz = 90 }\n  ```\n\n- Example 2:\n\n  ```toml\n  vrr = { cursor-hz = \"none\" }\n  ```\n",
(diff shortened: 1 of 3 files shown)


mahkoh:
>Also considered explaining why VrrMode::ALWAYS might not make sense, but wasn't sure how to phrase that. To me at least it was clear I didn't want that, other than for testing, but I imagine it's not obvious. Although, always with a high enough cursor-hz might even be okish for terminal/browser usage.

Ideally people should be able to run with VRR enabled always. I'm not doing this myself so there might be bugs or hardware limitations. Changing the cursor-hz could be done depending on window rules, maybe.

PR #729, opened 2026-01-26
Revert pull request #708

khyperia:
As discussed in https://github.com/mahkoh/jay/pull/708 the behavior in that PR is against spec.

I think, but I'm not actually sure, that the behavior before 708 is _also_ against spec, though: when `delta` is not a multiple of `AXIS_120`,

> This event does not occur on its own, it is coupled with a wl_pointer.axis event that represents this axis value on a continuous scale.

is not upheld, due to rounding towards zero.

I don't have time right now to figure out a "proper" fix (maybe sometime in the next few days), but, here's a revert!

### DIFF
--- a/src/ifs/wl_seat/event_handling.rs
+++ b/src/ifs/wl_seat/event_handling.rs
@@ -1459,11 +1459,7 @@ impl WlSeatGlobal {
                     if p.seat.version >= AXIS_VALUE120_SINCE_VERSION {
                         p.send_axis_value120(axis, delta);
                     } else if p.seat.version >= AXIS_DISCRETE_SINCE_VERSION {
-                        let mut accumulator = p.v120_accumulator[i].get();
-                        accumulator += delta;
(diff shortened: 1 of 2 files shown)


khyperia:
> I don't have time right now to figure out a "proper" fix (maybe sometime in the next few days), but, here's a revert!

Looks like wlroots does not send *any* events if its accumulator (the same accumulator I added in the previous PR in Jay) is less than a full scroll tick:

https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/436bcf9a4755b24f7fadc100ba8deaa7f343ba86/types/seat/wlr_seat_pointer.c#L352-358

Seems like reasonable behavior to me. If a client wants high resolution events, they should update to a wl_seat version with support for value120.

(I'm not sure how much worth there is in implementing fancy behavior for wl_seats `5 <= version < 8` though, it'll basically be dead code eventually... now that wine is on wl_seat version 8, I don't have much motivation to fix it, haha)

PR #746, opened 2026-02-17
config: add XWayland enabled option

khyperia:
I don't use any apps that require XWayland, and don't have XWayland installed. Jay still sets DISPLAY and attempts to exec XWayland and fails every time a client tries to connect. Add an option to disable it at startup! :sparkles:

For comparison against other compositors: [Hyprland supports disabling XWayland at runtime even if already started](https://github.com/hyprwm/Hyprland/blob/661314e13487784c94b3c9fd69b469764eb6ef7b/src/xwayland/XWayland.cpp#L10-L16). [Sway does not](https://github.com/swaywm/sway/blob/fa497964fd55632beacf5f425e964ae4893e25b9/sway/commands/xwayland.c#L26-L27).

### DIFF
--- a/src/config/handler.rs
+++ b/src/config/handler.rs
@@ -979,6 +979,11 @@ impl ConfigProxyHandler {
         Ok(())
     }
 
+    fn handle_set_x_wayland_enabled(&self, enabled: bool) -> Result<(), CphError> {
+        self.state.set_xwayland_enabled(enabled);
+        Ok(())
+    }
+
     fn handle_set_ui_drag_enabled(&self, enabled: bool) {
         self.state.ui_drag_enabled.set(enabled);
     }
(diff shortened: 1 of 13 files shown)


src/config/handler.rs
```diff
@@ -979,6 +979,18 @@ impl ConfigProxyHandler {
...
+    fn handle_set_x_wayland_enabled(&self, enabled: bool) -> Result<(), CphError> {
+        self.state.xwayland.enabled.set(enabled);
+        if enabled == self.state.xwayland.handler.borrow().is_some() {
+            Ok(())
+        } else if enabled {
+            self.state.start_xwayland();
+            Ok(())
+        } else {
+            Err(CphError::XWaylandAlreadyStarted)
+        }
```
mahkoh: This logic should be changed as follows:

- Add a boolean to XWaylandState that says whether xwayland is running.
- In xwayland::manage, set the boolean to true immediately before calling `run` and set it to false immediately after `run` returns.
- In xwayland::manage, add a new `on_drop` value to forker.unsetenv
- Add a function State::stop_xwayland that does the following:
  - Return immediately if xwayland is running
  - Otherwise drop the handler.

mahkoh (approved):

mahkoh:
Thanks!

I've made some small changes myself. A nice addition would be a cli command along the lines of `jay xwayland pid` that could be used to easily kill xwayland if it is running.

PR #785, opened 2026-03-10
Add mouse-refocus option to center cursor on focused window

Ktrompfl:
I don't know if there is a way to add this to the existing PR (https://github.com/mahkoh/jay/pull/665), so here is a new one. This hopefully implements what you suggested in https://github.com/mahkoh/jay/pull/665#issuecomment-4017035309.

The option mouse-refocus is now disabled by default, as it is a breaking change to the current behavior.
When enabled, focus changes via keyboard commands (move-left, focus-right, show-workspace, ...) asynchronously move the pointer to the center of the focused top-level node. In particular, if the focused is already focused, the pointer is still moved to the center.

There are a lot of different ways this could be implemented, and some other compositors have multiple options to specify the exact behavior or support features like persisting and restoring the relative pointer position per top-level node. For now, this implementation should at least cover the basic use cases. 

Closes #665

### DIFF
--- a/jay-config/src/_private/ipc.rs
+++ b/jay-config/src/_private/ipc.rs
@@ -871,6 +871,10 @@ pub enum ClientMessage<'a> {
         dx2: i32,
         dy2: i32,
     },
+    SeatSetMouseFollowsFocus {
+        seat: Seat,
+        enabled: bool,
+    },
 }
 
 #[derive(Serialize, Deserialize, Debug)]
--- a/src/config/handler.rs
+++ b/src/config/handler.rs
@@ -2425,6 +2425,16 @@ impl ConfigProxyHandler {
         Ok(())
     }
 
+    fn handle_seat_set_mouse_follows_focus(
+        &self,
+        seat: Seat,
+        enabled: bool,
+    ) -> Result<(), CphError> {
+        let seat = self.get_seat(seat)?;
+        seat.set_mouse_follows_focus(enabled);
+        Ok(())
+    }
+
     fn get_sized(&self, sized: Resizable) -> Result<ThemeSized, CphError> {
         use jay_config::theme::sized::*;
         let sized = match sized {
@@ -2597,6 +2607,7 @@ impl ConfigProxyHandler {
             return Err(CphError::WindowNotVisible(window_id));
         }
         seat.focus_toplevel(window);
+        seat.maybe_schedule_warp_mouse_to_focus();
         Ok(())
     }
 
@@ -3354,6 +3365,9 @@ impl ConfigProxyHandler {
             ClientMessage::SeatWarpMouseToFocus { seat } => self
                 .handle_seat_warp_mouse_to_focus(seat)
                 .wrn("seat_warp_mouse_to_focus")?,
+            ClientMessage::SeatSetMouseFollowsFocus { seat, enabled } => self
+                .handle_seat_set_mouse_follows_focus(seat, enabled)
+                .wrn("seat_set_mouse_follows_focus")?,
             ClientMessage::ConnectorSetUseNativeGamut {
                 connector,
                 use_native_gamut,
--- a/src/ifs/wl_seat.rs
+++ b/src/ifs/wl_seat.rs
@@ -260,6 +260,8 @@ pub struct WlSeatGlobal {
     simple_im: CloneCell<Option<Rc<SimpleIm>>>,
     simple_im_enabled: Cell<bool>,
     warp_mouse_to_focus_scheduled: Cell<bool>,
+    warp_mouse_to_focus_skip_target_check: Cell<bool>,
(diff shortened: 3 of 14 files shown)


jay-config/src/_private/ipc.rs
```diff
@@ -817,6 +817,10 @@ pub enum ClientMessage<'a> {
         position: BarPosition,
     },
     GetBarPosition,
+    SeatSetMouseRefocus {
+        seat: Seat,
+        enabled: bool,
+    },
```
mahkoh: Move this to the end of the enum.

src/ifs/wl_seat.rs
```diff
@@ -520,12 +524,16 @@ impl WlSeatGlobal {
...
 
-    pub fn set_workspace(&self, ws: &Rc<WorkspaceNode>) {
+    pub fn set_workspace(self: &Rc<Self>, ws: &Rc<WorkspaceNode>) {
         let tl = match self.keyboard_node.get().node_toplevel() {
             Some(tl) => tl,
             _ => return,
         };
         toplevel_set_workspace(&self.state, tl, ws);
+
+        if self.mouse_refocus.get() {
```
mahkoh: Move this check into schedule_mouse_refocus.

toml-config/src/config/parsers/config.rs
```diff
@@ -208,6 +209,7 @@ impl Parser for ConfigParser<'_> {
                 opt(val("simple-im")),
                 recover(opt(bol("show-titles"))),
                 opt(val("fallback-output-mode")),
+                recover(opt(bol("mouse-refocus"))),
```
mahkoh: This needs to be documented in spec.yaml.

Ktrompfl: Are the generated files created by CI or do I need to run `toml-spec` locally?

mahkoh: You need to run it locally.

Ktrompfl:
Alright, so now the renamed `mouse-follows-focus` option should try to warp the cursor to the center of the focused window whenever the user invokes an action that changes the focused window (hopefully all such actions are included now). This includes changing focus to the already focused window, in which case the cursor is merely re-centered. 

For consistency, this now excludes warping the cursor when the focused window is moved. To warp the cursor in such cases anyway, I've added a `warp-mouse-to-focus` action. 

For a more layout-aware approach, I've considered something like a feature with which the relative position of the pointer to a window is persisted and restored across all focus and layout changes and can only be changed with actual mouse movements. However, this would be considerably more complex, and I’m not convinced it would be especially useful. Therefore, I’m leaving it out of the scope of this pull request.

src/ifs/wl_seat.rs
```diff
@@ -818,6 +825,14 @@ impl WlSeatGlobal {
...
         }
+
+        self.schedule_warp_mouse_to_focus();
+    }
+
+    pub fn schedule_warp_mouse_to_focus(self: &Rc<Self>) {
+        if self.mouse_follows_focus() && !self.warp_mouse_to_focus_scheduled.replace(true) {
+            self.state.pending_warp_mouse_to_focus.push(self.clone());
+        }
     }
```
mahkoh, line 845: Split this up into `maybe_schedule_warp_mouse_to_focus` which checks for the config and `schedule_warp_mouse_to_focus` which does the rest.

src/config/handler.rs
```diff
@@ -2382,6 +2382,18 @@ impl ConfigProxyHandler {
...
 
+    fn handle_seat_set_mouse_follows_focus(&self, seat: Seat, enabled: bool) -> Result<(), CphError> {
+        let seat = self.get_seat(seat)?;
+        seat.set_mouse_follows_focus(enabled);
+        Ok(())
+    }
+
+     pub fn handle_seat_warp_mouse_to_focus(&self, seat: Seat) -> Result<(), CphError> {
+        let seat = self.get_seat(seat)?;
+        seat.schedule_warp_mouse_to_focus();
```
mahkoh: This needs to skip the check if mouse-follow-focus is enabled.

src/ifs/wl_seat.rs
```diff
@@ -1498,6 +1520,14 @@ impl WlSeatGlobal {
         self.focus_follows_mouse.get()
     }
 
+    pub fn set_mouse_follows_focus(&self, enabled: bool) {
+        self.mouse_follows_focus.set(enabled);
+    }
+
+    pub fn mouse_follows_focus(&self) -> bool {
+        self.mouse_follows_focus.get()
+    }
```
mahkoh, line 1540: This needs to be integrated with the control center. Look at some of the other setters here and see how they are used.

mahkoh (dismissed):

mahkoh (approved):

PR #804, opened 2026-03-15
control-center: close with escape

kotarac:
The new control-center is great, kudos again!

In case you wouldn't object to something like this, here's a PR and a couple of observations below.

`Esc` seemed to make most sense as I consider control-center a transient window that should be easy to close.
Considered `q`, `alt+f4`, `ctrl+w`, but preferred `Esc`.

`had_focus_prev_frame` and `had_popup_open_prev_frame` are to prevent `Esc` closing the control-center when a `TextEdit` is focused or a `ComboBox` has a selection popup open.
Widgets don't `consume_key` when clearing focus on `Esc` and I haven't found a more elegant way to solve this than passing information between frames.

### DIFF
--- a/src/control_center.rs
+++ b/src/control_center.rs
@@ -108,6 +108,8 @@ struct ControlCenterInner {
     window: Rc<EggWindow>,
     pane_ids: PaneIds,
     interests: Rc<Interests>,
+    had_focus_prev_frame: NumCell<bool>,
+    had_popup_open_prev_frame: NumCell<bool>,
 }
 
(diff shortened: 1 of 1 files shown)


mahkoh:
Are you using the control center primarily with the keyboard?

kotarac:
> If you're using it primarily with the mouse, then there is a close button in the top left that closes the window.
> 
> I'm hesitant to commit to escape closing the window.

That's totally ok! Will close this one.

PR #813, opened 2026-03-19
Add clean-logs-older-than-days option

khyperia:
My jay log directory is quite cluttered and ever-growing.

I don't know how to handle the portal's logs, as it's kind of buried in several forked processes deep... perhaps just hardcoding `../portal`. I figured I'd ask before attempting to implement something, in case you have ideas!

### DIFF
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -4,7 +4,7 @@ use {
         utils::{atomic_enum::AtomicEnum, errorfmt::ErrorFmt, oserror::OsError},
     },
     backtrace::Backtrace,
-    bstr::BString,
+    bstr::{BStr, BString, ByteSlice},
     log::{LevelFilter, Log, Metadata, Record},
     parking_lot::Mutex,
     std::{
@@ -17,9 +17,11 @@ use {
             Arc,
             atomic::{AtomicI32, AtomicU32, Ordering::Relaxed},
         },
+        thread,
         time::SystemTime,
     },
-    uapi::{Errno, Fd, OwnedFd, Ustring, c, format_ustr},
+    thiserror::Error,
+    uapi::{AsUstr, Dirent, Errno, Fd, OwnedFd, Ustring, c, format_ustr},
 };
 
 thread_local! {
@@ -80,6 +82,17 @@ impl Logger {
         log::set_max_level(filter);
     }
 
+    pub fn clean_logs_older_than(&self, time: SystemTime) {
+        let time_formatted = humantime::format_rfc3339_millis(time);
+        log::info!("Cleaning unused log files older than {}", time_formatted);
+        let path = self.path();
+        thread::spawn(move || {
+            if let Err(e) = clean_logs_older_than(path.as_bstr(), time) {
+                log::error!("Could not clean log files: {}", ErrorFmt(e));
+            }
+        });
+    }
+
     pub fn level(&self) -> LogLevel {
         self.level.load(Relaxed)
     }
@@ -105,6 +118,7 @@ impl Logger {
 
 pub fn open_log_file(ty: &str) -> (Ustring, OwnedFd) {
     let log_dir = create_log_dir(ty);
+    let mut flock_fail_count = 0;
     for i in 0.. {
         let file_name = format_ustr!(
             "{}/{ty}-{}-{}.txt",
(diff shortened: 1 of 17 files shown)


toml-config/src/config/parsers/config.rs
```diff
@@ -109,6 +111,7 @@ impl Parser for ConfigParser<'_> {
             ),
             (
                 log_level_val,
+                clean_logs_older_than_days_val,
```
mahkoh: Please add this at the end.

toml-spec/spec/spec.yaml
```diff
@@ -2624,6 +2624,18 @@ Config:
...
+      required: false
+      description: |
+        If specified on startup, deletes Jay's log files older than the
+        specified number of days.
+
+        - Example:
+
+          ```toml
+          clean-logs-older-than-days = 7
+          ```
```
mahkoh: Please make clean-logs-older-than a table with fields `days` and `weeks`.

src/logger.rs
```diff
@@ -80,6 +80,48 @@ impl Logger {
...
+            let mtime = metadata.modified()?;
+            if time.duration_since(mtime).is_ok() {
+                let entry_path = entry.path();
+                log::info!("Removing {entry_path:?}, as it is older than {time_formatted}");
+                std::fs::remove_file(entry_path)?;
+            }
+            Ok(())
+        };
+
+        if let Some(parent) = current_log_path.parent() {
```
mahkoh:
```suggestion
        let Some(parent) = current_log_path.parent() else {
```

src/logger.rs
```diff
@@ -80,6 +80,48 @@ impl Logger {
...
+            match parent.read_dir() {
+                Ok(entries) => {
+                    for entry in entries {
+                        match process_entry(entry) {
```
mahkoh:
```suggestion
                        if let Err(e) = process_entry(entry) {
```

src/logger.rs
```diff
@@ -80,6 +80,48 @@ impl Logger {
...
+            let entry = entry?;
+            if Some(entry.file_name()).as_deref() == current_log_path.file_name() {
+                return Ok(());
+            }
+            let metadata = entry.metadata()?;
+            if !metadata.is_file() {
+                return Ok(());
+            }
+            let mtime = metadata.modified()?;
+            if time.duration_since(mtime).is_ok() {
```
mahkoh: I don't want this to ever delete any files that are still in use even if they have not been modified. You can use `uapi::flock` for this purpose. See acceptor.rs.

src/logger.rs
```diff
@@ -209,3 +228,72 @@ impl Log for LogWrapper {
...
+            .map_err(Into::into)
+            .map_err(CleanLogsError::OpenFile)?;
+        let stat = uapi::fstat(*file)
+            .map_err(Into::into)
+            .map_err(CleanLogsError::Stat)?;
+        if stat.st_mtime >= time {
+            return Ok(());
+        }
+        if uapi::flock(file.raw(), c::LOCK_EX | c::LOCK_NB).is_err() {
+            return Ok(());
```
khyperia, line 301: > I've made some changes to the deletion logic. Please check if this works for you.

Personally I would prefer this to log something (whether that's an info! or warn! or error!, shrug) - it could be confusing if you have another Jay running on another VT and forget about it, or the file is flocked for some other reason, the file is silently kept around. Just a little nit though, totally works for my use case as-is. Otherwise everything works great from my testing just now!

mahkoh: You can add something like `Preserving file still in use:` if you want.

src/logger.rs
```diff
@@ -117,7 +130,13 @@ pub fn open_log_file(ty: &str) -> (Ustring, OwnedFd) {
             c::O_CREAT | c::O_EXCL | c::O_CLOEXEC | c::O_WRONLY,
             0o644,
         ) {
-            Ok(f) => return (file_name, f),
+            Ok(f) => {
+                if let Err(e) = uapi::flock(f.raw(), c::LOCK_EX | c::LOCK_NB) {
+                    log::warn!("Unable to flock just-opened logfile: {}", ErrorFmt(e));
+                    continue;
+                }
```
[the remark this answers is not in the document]
mahkoh: Please also address this. I think just removing the continue is ok.

toml-spec/spec/spec.yaml
```diff
@@ -3218,6 +3231,31 @@ GracePeriod:
       required: false
 
 
+CleanLogsOlderThan:
+  kind: table
+  description: |
+    The definition of how old logfiles need to be for them to be automatically deleted.
+    
+    Omitted values are set to 0.
```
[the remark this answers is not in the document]
mahkoh: I think the suggestion can be used as is.

src/logger.rs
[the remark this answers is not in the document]
khyperia: kind of wish `open()` had a `O_FLOCK_EX`, similar to how it has `O_CLOEXEC` to prevent race conditions and whatnot :P (but understandable that it doesn't, `O_CLOEXEC` solves completely unavoidable race conditions, but open-flock races are able to be detected and worked around)

src/logger.rs
```diff
@@ -117,7 +131,22 @@ pub fn open_log_file(ty: &str) -> (Ustring, OwnedFd) {
...
+            Ok(f) => {
+                if let Err(e) = uapi::flock(f.raw(), c::LOCK_EX | c::LOCK_NB) {
+                    log::warn!("Unable to flock just-opened logfile: {}", ErrorFmt(e));
+                    flock_fail_count += 1;
+                    if flock_fail_count > 10 {
+                        log::error!(
+                            "Failed to flock just-opened logfile more than 10 times in a row. Not
+                            flocking the logfile, if the cleanup routine later succeeds to flock
+                            this logfile, it will be deleted even if it is still in use."
+                        );
```
mahkoh:
```suggestion
                        log::error!(concat!(
                            "Failed to flock just-opened logfile more than 10 times in a row. Not ",
                            "flocking the logfile, if the cleanup routine later succeeds to flock ",
                            "this logfile, it will be deleted even if it is still in use.",
                        ));
```

khyperia: sorry, late response, was sick, haha - pushed now! thanks again for all the great work you do, Jay is such a breath of fresh air :sparkles:

mahkoh:
Thanks!

PR #816, opened 2026-03-19
config: workspace empty behavior

### LINKED ISSUE #519: workspace bar enhancements

kotarac:
Implementation according to our discussion here #541.
Closes #519.


mahkoh:
If you only have one display you can manually test this using virtual outputs and wl-mirror or obs.

kotarac:
> If you only have one display you can manually test this using virtual outputs and wl-mirror or obs.

Got one now actually, just need to get around to it. Rebasing it to stay up to date as I'm dogfooding this. :slightly_smiling_face:

kotarac:
Needs a bit more work, couple of edge cases to go through and some more `ext_workspace_manager_v1` work which has 0 test coverage and I don't usually use it.

mahkoh:
Since the toml config now has a workspaces table, this should also be configurable per-workspace.

kotarac:
> Since the toml config now has a workspaces table, this should also be configurable per-workspace.

Good point. Hadn't thought of that when I was rebasing it.

Think it should go with these changes or shall we do it later? It's already a pretty fat set of changes.

mahkoh:
>Think it should go with these changes or shall we do it later?

I hope you haven't been waiting for me to review this PR since I've been waiting for you to mark it as ready for review.

In any case, I think this PR should include a per-workspace setting.

kotarac:
> > Think it should go with these changes or shall we do it later?
> 
> I hope you haven't been waiting for me to review this PR since I've been waiting for you to mark it as ready for review.

No no. It's draft for a reason. It needed quite a bit of additional testing since the workspace protocol additions along with all the other stuff previously added. No worries. :slightly_smiling_face: 

> In any case, I think this PR should include a per-workspace setting.

Alright. I'll have a look.

kotarac:
Didn't feature gate IPC additions. We didn't do it for some other things which seem comparable, but wouldn't be opposed to adding it.

mahkoh:
Feature detection should only be used if not doing so would regress existing functionality.

kotarac:
Otherwise, tested the non-protocol paths quite a bit myself and have ran this for longer than the PR has been up, although almost exclusively set to `destroy` while not testing the work itself.

For the protocol changes, I tested a bit with waybar (I don't usually use it), but not nearly as extensively as with the above.

kotarac:
image

-.-

jay-config/src/lib.rs
```diff
@@ -223,6 +225,21 @@ impl Workspace {
     pub fn hide(self) {
         get!().hide_workspace(self);
     }
+
+    /// Sets what should happen when this workspace is empty.
+    pub fn set_empty_behavior(self, behavior: workspace::WorkspaceEmptyBehavior) {
```
mahkoh:
```suggestion
    pub fn set_empty_behavior(self, behavior: WorkspaceEmptyBehavior) {
```

jay-config/src/workspace.rs
```diff
@@ -17,3 +17,25 @@ pub enum WorkspaceDisplayOrder {
 pub fn set_workspace_display_order(order: WorkspaceDisplayOrder) {
     get!().set_workspace_display_order(order);
 }
+
+/// Configures what happens to empty workspaces when they are left or become inactive.
+#[derive(Serialize, Deserialize, Copy, Clone, Debug, Hash, Eq, PartialEq)]
+pub enum WorkspaceEmptyBehavior {
```
mahkoh, line 25:
```suggestion
#[derive(Serialize, Deserialize, Copy, Clone, Debug, Hash, Eq, PartialEq)]
#[non_exhaustive]
pub enum WorkspaceEmptyBehavior {
```

src/config/handler.rs
```diff
@@ -1251,7 +1266,8 @@ impl ConfigProxyHandler {
                 ws.desired_output.set(output.global.output_id.clone());
             }
             self.state.tree_changed();
-        } else {
+        }
+        if !move_ {
```
mahkoh: This is the same as

```rust
if move_ && !ws.hidden.get() {
```

and everything else unchanged.

src/config/handler.rs
```diff
@@ -1265,6 +1281,9 @@ impl ConfigProxyHandler {
     fn handle_set_seat_workspace(&self, seat: Seat, ws: Workspace) -> Result<(), CphError> {
         let seat = self.get_seat(seat)?;
         let ws = self.get_workspace(ws)?;
+        if seat.get_keyboard_node().node_toplevel().is_none() {
+            return Ok(());
+        }
```
mahkoh: If you're fetching the node anyway pass it into set_workspace.

src/config/handler.rs
```diff
@@ -1274,6 +1293,14 @@ impl ConfigProxyHandler {
                 WorkspaceType::Overlay => self.state.create_overlay_workspace(&ws.name),
             },
         };
+        if workspace.hidden.get()
+            && self
+                .state
+                .restore_hidden_workspace(&workspace, Some(seat.get_fallback_output()), Some(&seat))
```
mahkoh: This seems to contradict the documented behavior of hidden workspaces

```
                                           When restoring, the compositor
prefers the output the workspace was last shown on if it is still connected.
```

src/config/handler.rs
```diff
@@ -1291,6 +1318,14 @@ impl ConfigProxyHandler {
                 WorkspaceType::Overlay => self.state.create_overlay_workspace(&ws.name),
             },
         };
+        if workspace.hidden.get()
+            && self
+                .state
+                .restore_hidden_workspace(&workspace, window.node_output(), None)
```
mahkoh: Same here.

src/ifs/workspace_manager/ext_workspace_manager_v1.rs
```diff
@@ -256,24 +264,32 @@ impl ExtWorkspaceManagerV1RequestHandler for ExtWorkspaceManagerV1 {
...
-                    };
-                    move_ws_to_output(&ws, &o, config);
-                    ws.desired_output.set(o.global.output_id.clone());
-                    self.client.state.tree_changed();
+                    self.assign_workspace(&ws, &o);
                 }
                 WorkspaceChange::CreateWorkspace(name, output) => {
                     if self.client.state.workspaces.contains(&name) {
-                        return Ok(());
+                        continue;
```
mahkoh, line 267: This looks like a bug fix that does not belong in this commit.

src/tree/float.rs
mahkoh:
```suggestion
        let old_ws = self.workspace.set(ws.clone());
```

src/tree/output.rs
```diff
@@ -936,7 +940,23 @@ impl OutputNode {
...
+        self: &Rc<Self>,
+        name: &str,
+    ) -> Rc<WorkspaceNode> {
+        self.create_workspace2(name, WorkspaceType::Normal, false)
+    }
+
+    fn create_workspace2(
+        self: &Rc<Self>,
+        name: &str,
+        ty: WorkspaceType,
```
mahkoh:
```suggestion
```

src/tree/workspace.rs
```diff
@@ -218,6 +229,16 @@ impl WorkspaceNode {
         self.state.trigger_cci(CCI_WORKSPACES);
     }
 
+    pub fn set_hidden_output(&self, output: &Rc<OutputNode>) {
+        if !self.hidden.get() {
+            return;
+        }
+        if self.output.get().id == output.id {
```
mahkoh:
```suggestion
        if self.output.id() == output.id {
```

src/tree/workspace.rs
```diff
@@ -635,8 +662,13 @@ pub fn move_ws_to_output(ws: &Rc<WorkspaceNode>, target: &Rc<OutputNode>, config
             || (config.make_visible_if_empty && target.workspace.is_none()));
     if make_visible {
         ws.state.show_workspace2(None, target, &ws);
-    } else {
+    }
+    if !make_visible {
```
mahkoh: What is the point of this?

src/state.rs
```diff
@@ -873,21 +874,28 @@ impl State {
                 return Some(on()?.ensure_normal_workspace());
             };
             let ty = s.workspace_ty.get().unwrap_or(WorkspaceType::Normal);
+            let restore = |ws: Rc<WorkspaceNode>, output| {
+                if !ws.hidden.get() {
+                    return Some(ws);
+                }
+                self.restore_hidden_workspace(&ws, output, None)?;
```
mahkoh:
```suggestion
                if ws.hidden.get() {
                    self.restore_hidden_workspace(&ws, output, None)?;
                }
```

src/state.rs
```diff
@@ -1067,14 +1084,99 @@ impl State {
         did_focus
     }
 
+    pub fn restore_hidden_workspace(
+        &self,
+        ws: &Rc<WorkspaceNode>,
```
mahkoh: All of these function that take WorkspaceNode as an argument should be defined on WorkspaceNode.

src/state.rs
```diff
@@ -1849,7 +1960,12 @@ impl State {
         if output.is_dummy {
             return;
         }
-        if ws.output.id() == output.id {
+        ws.desired_output.set(output.global.output_id.clone());
+        if ws.output.get().id == output.id {
```
mahkoh:
```suggestion
        if ws.output.id() == output.id {
```

src/state.rs
```diff
@@ -1972,6 +2087,135 @@ impl State {
...
+        }
+        for (client, e) in clients_to_kill.values() {
+            client.error(e);
+        }
+    }
+
+    pub fn set_workspace_empty_behavior(&self, behavior: WorkspaceEmptyBehavior) {
+        self.workspace_empty_behavior.set(behavior);
+        self.trigger_cci(CCI_COMPOSITOR);
+        if !matches!(
```
mahkoh:
```suggestion
        if not_matches!(
```

src/state.rs
```diff
@@ -1972,6 +2087,135 @@ impl State {
...
+            WorkspaceEmptyBehavior::DestroyOnLeave => {}
+            WorkspaceEmptyBehavior::HideOnLeave => {}
+            WorkspaceEmptyBehavior::Destroy => self.destroy_empty_workspace(ws),
+            WorkspaceEmptyBehavior::Hide => self.hide_empty_workspace(ws, update_desired_output),
+        }
+    }
+
+    fn workspace_is_active(&self, ws: &WorkspaceNode) -> bool {
+        let output = ws.output.get();
+        if let Some(active) = output.workspace.get() {
```
mahkoh:
```suggestion
        if let Some(active) = output.workspace.id() {
```

mahkoh (changes requested):
I feel like hidden workspaces should stay attached to their output. That would also remove the need to find the correct output when unhiding the workspace and would keep the workspace in the same position as before.

Sometimes it might be useful to have a linked list containing only the visible workspaces. For such cases it might be useful to maintain two linked lists per output. One containing all workspaces and one containing only the visible workspaces.

kotarac:
> I feel like hidden workspaces should stay attached to their output. That would also remove the need to find the correct output when unhiding the workspace and would keep the workspace in the same position as before.

Hmm, that would help. And yes, the current approach didn't place it back where it was in case the display order was manual.
 
> Sometimes it might be useful to have a linked list containing only the visible workspaces. For such cases it might be useful to maintain two linked lists per output. One containing all workspaces and one containing only the visible workspaces.

:+1:

src/ifs/workspace_manager/ext_workspace_manager_v1.rs
kotarac: Will submit a separate PR.

jay-config/src/workspace.rs
kotarac: What do you think we might want to add here in the future? Or leaving it open just in case?

src/tree/workspace.rs
kotarac: Style, I have a habbit of killing `else` blocks that's often inconsistent with the rest here. :slightly_smiling_face: 

Will get it reverted.

kotarac:
Think the separate linked list only for visible workspaces is needed? Seems cheap to just filter out hidden ones where needed, but it is inconvenient that it's done in multiple places though...

src/state.rs
mahkoh: I meant for this to apply to all of the functions in this file.

src/tree/output.rs
```diff
@@ -2134,11 +2153,22 @@ impl Node for OutputNode {
...
+                };
+                let Some(new) = new else {
+                    break;
+                };
+                candidate = new;
+                if !candidate.hidden.get() {
+                    ws = candidate;
+                    break;
+                }
+            }
```
mahkoh: Add additional functions to NodeRef that allow walking the chain until a condition is true.

src/state.rs
mahkoh: I only meant those functions added by this PR.

mahkoh: I only meant those functions added by this PR.

src/state.rs
```diff
@@ -984,22 +991,22 @@ impl State {
...
+        let last_active_parent = last_active
+            .tl_data()
+            .parent
+            .get()
+            .and_then(|node| node.node_into_container());
+        let Some(last_active_parent) = last_active_parent else {
+            container.append_child(node);
+            return;
+        };
+        last_active_parent.add_child_after(&*last_active, node);
```
mahkoh: I have no idea what these changes are doing. If such refactorings were ok, and I'm not saying that they are, then they don't belong in the same commit as feature additions. Changes to behavior and changes that keep the behavior the same should be in separate commits if possible.

kotarac: Sorry, the back and forth on the src/state.rs fns taking WorkspaceNode had me mess up things quite a bit.

kotarac:
FTR since the last review, I discovered a small issue when `destroy` is set with floating windows, that's another extra change since.

src/state.rs
```diff
@@ -1101,7 +1116,7 @@ impl State {
         ty: WorkspaceType,
         mut output: Option<Rc<OutputNode>>,
     ) {
-        let mut output = || {
+        let mut resolve_output = || {
```
kotarac, line 1272: Might make sense to extract to a helper fn.

src/tree/toplevel.rs
```diff
@@ -1086,13 +1086,19 @@ pub fn toplevel_set_floating(state: &Rc<State>, tl: Rc<dyn ToplevelNode>, floati
...
-        let (width, height) = data.float_size(&ws);
-        state.map_floating(tl, width, height, &ws, None);
+        return;
     }
+    let Some(ws) = data.workspace.get() else {
+        return;
+    };
+    let (width, height) = data.float_size(&ws);
+    state.map_floating(tl.clone(), width, height, &ws, None);
+    parent.cnode_remove_child2(&*tl, true);
```
mahkoh: What does this do?

kotarac: https://github.com/mahkoh/jay/pull/816#issuecomment-4415259421

Single window remaining on a workspace with empty behavior set to destroy being toggled between tiled/floating.

mahkoh: This seems to change the behavior of toggling a window from floating to tiled.

kotarac: Will take a look ~wednesday. Right now, looking at this on the phone and answering by memory. :slightly_smiling_face:

kotarac: It's wednesday. :smile:

kotarac:
Moved it to draft. It's been a while, a lot has changed, and I haven't had time/willpower to move it. Will see if I can find some time to test it properly again.

mahkoh:
Don't worry about it.

PR #821, opened 2026-03-20
config: add resize action

### LINKED ISSUE #637: Resize/Move methods on `Window` for floating windows (and beyond)
This is mentioned in other issues. But I want to request something specific for me as a `jay_config` user.

Resize/Move methods on `Window` would be great. Using the mouse with window management is a no go for me with floating windows.

Beyond that, resizing  tiling windows would be a great addition too (a la #574).

Relatedly, `Window::move_` is maybe not the best name to convey what the method does. Maybe `move_replace` or some other name would be more accurate/self-explanatory.

I'm liking `jay_config` a lot, and maybe it will help me move away from Awesome eventually. But those are important things that are still missing IMHO.

mahkoh:
Closes #637

### DIFF
--- a/book/AGENTS.md
+++ b/book/AGENTS.md
@@ -1,299 +1,135 @@
 # Jay Book -- Agent Instructions
 
-This file provides context for AI agents working on the Jay user documentation
-book (mdbook). Read this before making changes.
+User-facing [mdbook](https://rust-lang.github.io/mdBook/) documentation for
+the [Jay Wayland compositor](https://github.com/mahkoh/jay). Target audience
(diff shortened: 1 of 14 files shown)


wind0204:
Hi, I'm used to tiled window managers and I was wondering if I could configure jay so that jay would automatically choose an appropriate edge to move between `dx1` and `dx2`; e.g. choose dx1 if the right edge of the window is touching the right edge of the whole workspace/screen and choose dx2 otherwise.

It seems jay would just ignore the command inputs if the edge is not movable when the edge is touching the edge of the screen.

PR #826, opened 2026-03-22
cmm: Make log_100 and log_316 TFs symmetric

kennylevinsen:
log_100 and log_316 has the treshold implemented in the inverse EOTF, but missing from the EOTF. This causes black from a client buffer to be elevated to 0.01 and 0.00316 respectively, which can be seen in wl-colortest as having the black brightness tile become grey as you switch through the TFs.

Add the threshold to the EOTF so that 0 means black.


src/gfx_apis/vulkan/shaders/eotfs.glsl
```diff
@@ -67,7 +67,11 @@ vec3 inv_eotf_st240(vec3 c) {
 }
 
 vec3 eotf_log100(vec3 c) {
-    return pow(vec3(10), vec3(2.0) * (c - vec3(1.0)));
+    return mix(
+        vec3(0.0),
+        pow(vec3(10), vec3(2.0) * (c - vec3(1.0))),
+        greaterThan(c, vec3(0.0))
+    );
```
mahkoh, line 74: I feel like this should just `clamp(c, 0.0, 1.0)` like the other SDR transfer functions.

kennylevinsen: The problem is that when `c` is zero, `pow(vec3(10), vec3(2.0) * (c - vec3(1.0)))` yields `0.01`, so a clamp wouldn't help.

The approach is similar to `compound_power_2_4`. and matches what is done in their respective inverse EOTF's (a mix over `greaterThanEqual(c, vec3(0.01))` and `greaterThanEqual(c, vec3(sqrt(10)/1000))`, respectively).

mahkoh: The problem I see with your current code is that it creates a discontinuity at 0. I.e. `E=0` maps to `O=0` but `E=epsilon` maps to `O=0.01`. For a single black tile that would be fine, but I'm not sure if it would be fine for a gradient or any image that just so happens to have a black pixel.

kennylevinsen: The step in log_316 is small enough that it is unlikely to matter, but log100 clearly shows the discontinuity in the gradient (images below).

Searching around a bit I don't see a whole lot of implementations to reference, but the fix matches [gstreamer](https://github.com/GStreamer/gst-plugins-base/blob/ce937bcb21412d7b3539a2da0509cc96260562f8/gst-libs/gst/video/video-color.c#L609) and the [color-science python package](https://github.com/colour-science/colour/blob/c209b6c1e451b4daabcda015c611d4e628db204c/colour/models/rgb/transfer_functions/itut_h_273.py#L176). moxcms does neither and [instead uses the midpoint](https://github.com/awxkee/moxcms/blob/0a9c5d2d19808c7459c992aa47beb2c2d081047d/src/gamma.rs#L248) to minimize the maximum error. Not a whole lot of support out there for these TFs though, so don't really have much in way of a benchmark or convention...

Plain gamma22:
image

Before this fix:
image

After this fix (note the left side of the gradient bar):
image

mahkoh: If we want to use 0 as the fixed point, which is what all other EOTFs do right now, then maybe applying an affine function could work:

```glsl
vec3 eotf_log100(vec3 c) {
    c = clamp(c, 0.0, 1.0);
    c = pow(vec3(10), vec3(2.0) * (c - vec3(1.0)));
    return (c - vec3(0.01)) / vec3(1.0 - 0.01);
}

vec3 inv_eotf_log100(vec3 c) {
    c = clamp(c, 0.0, 1.0);
    c = c * vec3(1.0 - 0.01) + vec3(0.01);
    return vec3(1.0) + log2(c) / vec3(log2(10)) / vec3(2.0);
}
```

However, I'd prefer to standardize these things in https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/staging/color-management/appendix.md.

PR #830, opened 2026-03-23
cli: add json output

mahkoh:


src/cli.rs
```diff
@@ -53,6 +56,14 @@ pub struct GlobalArgs {
     /// The log level.
     #[clap(value_enum, long, default_value_t)]
     pub log_level: LogLevel,
+    /// Output data as json.
```
kotarac: perhaps should be clear it's specifically JSON lines

PR #868, opened 2026-04-10
Update Nix flake

Ktrompfl:
This PR makes the following changes to the recently added Nix flake:

- Adds missing build inputs that previously caused the package to fail to build with `nix build`.
- Adds the Nix lock file to the repository. This helps ensure reproducible builds and should likely be version-controlled.
- Installs shell completions and the Wayland session file as part of the package.
- Skips tests that require `io_uring` access, since this is not permitted in the sandboxed build environment. While this restriction does not appear to be always enforced by CppNix, it causes failures with other implementations such as Lix (see https://hydra.nixos.org/build/326198517/nixlog/1).

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -83,6 +83,15 @@
             vulkan-loader
           ];
 
+          checkFlags = [
+            # the following tests require access to io_uring, which is disabled in the sandboxed build environment
+            "--skip=cpu_worker::tests::cancel"
+            "--skip=cpu_worker::tests::complete"
+            "--skip=eventfd_cache::tests::test"
+            "--skip=io_uring::ops::read_write_no_cancel::tests::cancel_in_kernel"
+            "--skip=io_uring::ops::read_write_no_cancel::tests::cancel_in_userspace"
+          ];
+
           postInstall = ''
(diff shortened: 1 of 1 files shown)


mahkoh:
>Adds the Nix lock file to the repository. This helps ensure reproducible builds and should likely be version-controlled.

I'd prefer not to do this. Jay only supports the latest stable rust version and adding a lock file pins the rust version.

>Skips tests that require io_uring access, since this is not permitted in the sandboxed build environment. While this restriction does not appear to be always enforced by CppNix, it causes failures with other implementations such as Lix (see hydra.nixos.org/build/326198517/nixlog/1).

That's fine but I won't maintain that list.

Ktrompfl:
> I'd prefer not to do this. Jay only supports the latest stable rust version and adding a lock file pins the rust version.

Alright, makes sense. I've removed the lock file again.

> That's fine but I won't maintain that list.

Fair enough. If this ever expands beyond a few tests that can be maintained via occasional PRs, all tests could be disabled with `doCheck = false` to ensure the package remains buildable. For now, however, it’s probably best to keep the functional tests enabled.

mahkoh (approved):

mahkoh:
Thanks!

PR #884, opened 2026-04-16
it: data dir in testruns

kotarac:
While there's nothing that can go wrong atm, at least from what I can see, it seems weird that session management integration tests touch `$HOME/.local/share/jay/` and share locks (not that it would ever make a diff, but still).

This doesn't split it up per test run, but that doesn't really make a difference atm.

### DIFF
--- a/src/utils/data_dir.rs
+++ b/src/utils/data_dir.rs
@@ -3,6 +3,14 @@ use std::{
     sync::LazyLock,
 };
 
+#[cfg(feature = "it")]
+pub fn data_dir() -> &'static Path {
+    static DIR: LazyLock<PathBuf> =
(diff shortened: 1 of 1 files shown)


src/utils/data_dir.rs
```diff
@@ -4,12 +4,20 @@ use std::{
 };
 
 pub fn data_dir() -> &'static Path {
```
mahkoh, line 7:
```suggestion
#[cfg(not(feature = "it"))]
pub fn data_dir() -> &'static Path {
```

mahkoh:
Please put it under `./testruns` instead of `/tmp`.

kotarac:
Went with `testruns/data/`.

Still not split per testrun, don't think that's worth the plumbing changes needed.

mahkoh (approved):

mahkoh:
Thanks!

PR #888, opened 2026-04-19
config: support disabling outputs

marienz:
Useful for disabling something like a TV by default without having to keep track of which physical connector it is connected to.


mahkoh (approved):

mahkoh:
Thanks!

PR #895, opened 2026-04-25
opengl: use versioned SONAMEs for EGL and GLESv2

polycatenane:
This prevents runtime failure on Linux distributions where unversioned libraries are only included with the development packages.


mahkoh (approved):

mahkoh:
Thanks!

PR #932, opened 2026-05-09
ext-workspace: keep draining commit

kotarac:
If a workspace already exists or an output is stale, continue processing the commit.

### DIFF
--- a/src/ifs/workspace_manager/ext_workspace_manager_v1.rs
+++ b/src/ifs/workspace_manager/ext_workspace_manager_v1.rs
@@ -268,10 +268,10 @@ impl ExtWorkspaceManagerV1RequestHandler for ExtWorkspaceManagerV1 {
                 }
                 WorkspaceChange::CreateWorkspace(name, output) => {
                     if self.client.state.workspaces.contains(&name) {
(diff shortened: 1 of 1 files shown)


src/ifs/workspace_manager/ext_workspace_manager_v1.rs
```diff
@@ -268,7 +268,7 @@ impl ExtWorkspaceManagerV1RequestHandler for ExtWorkspaceManagerV1 {
                 }
                 WorkspaceChange::CreateWorkspace(name, output) => {
                     if self.client.state.workspaces.contains(&name) {
-                        return Ok(());
+                        continue;
                     }
                     let Some(output) = output.node() else {
                         return Ok(());
```
mahkoh: Why does this return?

kotarac: Well, I suppose this one shouldn't return either. But I think a user encountering this issue is much less likely than in the 1st case.

Will adjust.

mahkoh (approved):

mahkoh:
Thanks!

PR #1010, opened 2026-06-16
don't hold drm_devs lock while calling make_render_device in devices_enumerated

### LINKED ISSUE #1009: dma buffers are very sad on latest master
Image

flickers to a different pattern every alacritty redraw (e.g. upon input, or cursor blink), sometimes mostly black, sometimes only a few little streaks

bisected to this exact commit https://github.com/mahkoh/jay/commit/56629c39a16e106c922c22e905ed3edfed5e76e3

Running with `JAY_NO_CLIENT_PRIME=1` "fixes" the issue. I **only** have an nvidia 2080, intel integrated gpu disabled in bios, two monitors. I haven't poked around that commit yet to see what might be causing it, will hopefully have time to do so soon, but figured I'd post this in case anyone else might be having the same issue!

khyperia:
fixes https://github.com/mahkoh/jay/issues/1009

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -716,7 +716,8 @@ impl State {
             config.devices_enumerated()
         }
         if self.render_ctx.is_none() {
-            for dev in self.drm_devs.lock().values() {
+            let drm_devs: Vec<_> = self.drm_devs.lock().values().cloned().collect();
+            for dev in &drm_devs {
                 if let Ok(version) = dev.dev.version()
                     && version.name.contains_str("nvidia")
                 {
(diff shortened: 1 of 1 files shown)


mahkoh (approved):

mahkoh:
Let's go with the simpler version. CopyHashMaps are locked in many places.

The better solution would probably be a debug assertion when get or set is called while it is locked.

Please test this version.

khyperia:
sounds good - and the new version works great, thanks! :heart:

mahkoh:
Thanks!

PR #1021, opened 2026-06-21
Most frames are mostly perfect

mahkoh:
I used the following instructions to have claude review these patches for me. I don't think claude did a very good job, though.

---

# Background

## Jay Event Loop

Each iteration of the Jay event loop runs four stages:

1. Event Handling
2. Layout
3. Post Layout
4. Rendering

Each operation within a stage can schedule new operations. If the operation is
in the same stage, it will run before progressing to the next stage. Otherwise
it will run the next time the respective range is reached.

The Event Handling stage reacts to user input and client input and adjusts the
position and sizes of objects on screen. If this happens, it likely also
schedules work in the Layout stage to run.

The Layout stage takes such changes and performs layout for all affected
objects, propagating changes downwards in the on-screen tree.

The Post Layout stage performs tasks such as rendering new titles. For example,
if the width of a window has changed, the title of the window might need to be
ellipsized differently.

The Rendering stage performs rendering based on the current state of the tree.
Naturally, most event loop iterations do not perform rendering since it is only
performed once per refresh cycle.

## Wayland Configuration Sequences

Wayland uses configuration sequences to synchronize layout changes between the
client and the compositor. For example

1. The compositor wants the size of the window to be increased by 1px.
2. It sends a configuration sequence with a serial and the new size to the
   client.
3. The client prepares a new buffer for the requested size.
4. The client acks the serial and attaches the new buffer.
5. The client commits the pending changes.

The compositor CAN use this to delay changes on screen. For example, the
compositor might draw server-side decorations around windows. If the compositor
does not delay, then there will be a 1px gap between the window contents and the
decorations until the client commits. If the compositor does delay, then it will
delay the decoration update until the client acks and commits the requetsed
change.

This is known as "every frame is perfect".

Jay currently DOES NOT delay. When the Layout stage computes a new layout and
sends new configuration to a client, the Render stage picks up the new layout
immediately without waiting for the client to adjust its buffer.

## Synchronizing Between Multiple Clients

When synchronizing between multiple clients, care must be taken not to have more
than one configuration sequence in flight. For example, consider two clients A,
B being resized at once. Client A is shrinking and client B is growing by the
same amount.

If we allowed multiple configuration sequences in flight, we might send
configuration serials

  1, 2, 3, 4, 5, ...

to both clients without waiting for them to ack and commit.

In this case, client A might ack 1, 3, 5, ... and client B might ack 2, 4, 6, ...

There would therefore never be a state where both clients are in sync.
Therefore, no matter at what point we decided to render, client A's buffer would
be one pixel too large or client B's buffer would be one pixel too small.

This issue has already been solved in Jay. Jay uses configuration groups to
handle synchronization between multiple clients. Each group always sends its
configuration sequences to all of its members at once. Each group is blocked
until all of its members are idle, i.e. have no sequence in flight, i.e.
have acked and committed the previous sequence.

Jay uses a flood-fill algorithm to skip over groups that have successors that
only depend on their predecessors. This is an important optimization since, in
the example above, it allows us to jump from serial 1 to serial 10 to serial 20
and so on instead of having to roundtrip for each serial individually.

## Rendering in Jay

Rendering in Jay is controlled by the renderer.rs file. It walks the tree and
collects a flat list of instructions for the rendering backend. There are
currently only two kinds of instructions:

1. Fill a rectangle with a color.
2. Copy (parts of) a texture to a rectangle.

The rendering backend then executes these instructions.

To reduce the amount of work done by the GPU, Jay also performs damage tracking.
Damage is a set of rectangles that describe which parts of the screen have
changed. The rendering backend MAY skip rendering any pixel of the screen that
is not covered by a damage rectangle.

Damage does not affect the creation of the rendering instructions. It only takes
effect when the instructions are processed by the rendering backend.

Damage does, however, affect when rendering is triggered. If there is no damage,
then no rendering will be performed and therefore the list of instructions will
also never be created.

Various parts of Jay use the `State::damage` function to add damage.

Applying too much damage has no serious consequences as long as it doesn't
happen to often. In the worst case, it wastes some resources.

Applying to little damage can be fatal. For example, consider the case where a
client changes parts of its buffer but damage is not applied correctly. Then,
when the screen is rendered the next time, the renderer might skip rendering
the part of the screen where the buffer appears. Let's call the framebuffer that
was rendered to here FB1. If that area is damaged before the next rendering, the
renderer will render that part of the framebuffer, call it FB2. After that,
whenever the display switches between FB1 and FB2, that area of the screen will
flicker, until such a time when it is also draw over in FB1.

The fundamental rule for damage is therefore:

- Assume that the screen was just rendered into FB1. If we rendered again right
  now into FB2, then all of the pixels of FB2 that would be different from FB1
  must be damaged.

# The Goal

The goal of these changes is to make every frame perfect.

That is not fully possible in all situations:

- Not all surfaces types/surface changes use configuration sequences.
- There has to be some kind of timeout when waiting for the client to react to
  requested changes, otherwise the desktop would get stuck.

Therefore, it might be more accurate to say that "most frames are mostly
perfect".

# The Implementation

Since we want to make sure that instructions created by renderer.rs create a
"perfect" image, we have to look at which data the renderer uses to create the
instructions.

For all of that data, we create two timelines:

- The Live timeline: It encodes the imperfect, live state reflecting the latest
  changes.
- The Render timeline: It lags behind the Live timeline until all affected
  clients have adjusted themselves. Once that happens, a batch of changes are
  applied so that the data on the render timeline is once again perfect.

For all data that is such split, the renderer only uses the data from the Render
timeline.

For example, when increasing the width of a window from 500px to 501px, the live
timeline would immediately say that the width is 501px, but the render timeline
would wait until the client has adjusted its buffer to 501px before applying
that change.

All changes to the render timeline are organized in so-called transactions. In
the Event Handling and Layout stages, components can add themselves to the
transaction. In the Post Layout stage, a task runs that commits the transaction.
The transaction then includes all of the changes submitted by these components
plus all of the surfaces that had configuration sequences sent to them during
this event loop iteration.

This works because there should be no changes to the render state in the Post
Layout and Render stages.

A transactions dependencies are determined as follows:

- For each component that is part of a transaction, there is a dependency on the
  previous transaction of that component, so that all transactions of a
  component always execute in order.
- For each involved surface, there is a dependency waiting for that surface
  being ready to apply a commit that acks the transaction serial.

Application of surface commits containing serials is blocked on the transaction
containing the serial having been applied. This means that, for the surface
state that is controlled by the client, the live and the render timeline
coincide.

Each transaction has a timeout, by default 50 ms. If clients do not signal
readiness in time, the transaction is applied anyway. This means that the render
timeline lags behind the live timeline by at most 50 ms.

If this happens, the surfaces that weer too slow are marked as tardy. Tardy
surfaces do not block future transactions until they can prove that they once
again commit serials within the timeout, at which point the tardy mark is
removed.

# Concerns

## Damage Tracking

To ensure that we uphold the fundamental rule of damage, we must move some
damage applications to the render timeline. This is easy to get wrong and
therefore needs to be checked carefully.

## Leaks

Since transactions run asynchronously, it is possible to introduce leaks through
them. We must ensure that all such leaks are prevented.

## Missed Data

We might have missed some data used by the renderer.

## Violated Invariants

We might have violated invariants that cause the compositor to crash or
misbehave.

## Accepted Limitations

The following are known and accepted, not bugs:

- Input is hit-tested on the live timeline while the user sees the render
  timeline, so for up to the timeout clicks/focus resolve against unseen geometry.
- All surfaces configured in one event loop iteration share a transaction, so the
  slowest one delays buffer release/frame callbacks of the others until it commits
  or the timeout elapses.
- Deferred commit application shifts fifo/commit-timing/presentation-feedback
  timing, but only during reconfiguration, not at rest.
- A tardy surface is rendered with its old buffer at its new geometry.

# Commit Structure

The first few commits add the transaction infrastructure.

Then, a large number of commits duplicates data from the live timeline to the
render timeline.

Then, finally, a single commit switches rendering from the live timeline to the
render timeline.

# Q&A

Clarifications given during review:

Q: Why two separate timeout systems (configure-group vs transaction)?
A: Intentional. They are independent and can be tuned individually.

Q: The transaction timeout clock starts at commit time, not configure-send time,
   so a surface whose configure was merely queued can be marked tardy.
A: Acceptable. The timeout only governs the maximum lag of the render timeline.

Q: What is in/out of scope for the render-timeline split?
A: Output mode/scale/transform are out of scope (not configuration sequences).
   Subsurface relative positions are transactional. Cursor surfaces, layer-surface
   exclusive zones, and drag offsets are not transactional.

Q: The node_visible debug_assert that becomes a warning?
A: It should always be a warning (don't crash). It should never trigger; if it
   does, that's a serious bug. The assert-then-warning churn will be rebased away.

Q: When is damage applied?
A: All damage is applied on the render timeline, except where the live and render
   timelines collapse (e.g. when commits are applied), where it is immediate.
   Non-transactional fields are always collapsed.

Q: What happens to split fields of destroyed/disconnected clients?
A: All changes to split fields are applied on the render timeline except during
   shutdown. Client objects can remain part of the render timeline after the client
   disconnects; their buffers stay usable until removed on the render timeline.

Q: Can a transaction's contents change after commit?
A: No. Contents are fixed at commit time; only blockers for subsequent
   transactions may be added afterwards.

Q: An out-of-scope change can cause an in-scope change (e.g. output scale ->
   child geometry), splitting one action across timelines.
A: Acceptable limitation. The goal for now is that most frames are mostly perfect.

Q: Do the two timeouts need a particular relationship for correctness?
A: No configuration should break any invariant, though different values have not
   been experimented with.

Q: How do Xwayland surfaces synchronize, given they are transactional but carry no
   configure serial?
A: X surfaces don't use configuration sequences, so there is nothing to
   synchronize with. The xwayland_serial is an unrelated concept.

Q: Can fifo/mailbox cause a later commit to bypass a serial-blocked commit?
A: No. Commits are always applied in order.

Q: How do the integration tests stay deterministic under the async timeout?
A: The tests have not been looked at yet; ignore them for now.


mahkoh:
I had to remove part of the original post because it confused copilot. Use the history to see the original.

PR #1047, opened 2026-06-29
flake: update included fileset

Ktrompfl:
Sorry to open yet another PR for the Nix flake but `codegen/` was added recently and isn't included in the flake's source fileset, which breaks the build again. 

I've changed the derivation to now include all git-tracked files so future additions are picked up automatically.


mahkoh (approved):

mahkoh:
Thanks!

Issue #1052, opened 2026-06-30
Add action for manipulating floating window's absolute size and position


pattydaone:
The title is perhaps self explanatory. This would be most useful for window rules, where one may want a certain application to be of a certain size and position.

PR #1069, opened 2026-07-06
config: add absolute size / position getters / setters and window rules

Ktrompfl:
Implements https://github.com/mahkoh/jay/issues/1052.


jay-config/src/_private/ipc.rs
```diff
@@ -719,6 +719,16 @@ pub enum ClientMessage<'a> {
...
+    SetWindowMatcherInitialFloatingSize {
+        matcher: WindowMatcher,
+        width: i32,
+        height: i32,
+    },
+    SetWindowMatcherInitialFloatingPosition {
+        matcher: WindowMatcher,
+        x: i32,
+        y: i32,
+    },
```

jay-config/src/_private/ipc.rs
```diff
@@ -881,6 +891,28 @@ pub enum ClientMessage<'a> {
...
+        window: Window,
+        x: i32,
+        y: i32,
+    },
+    GetWorkspacePosition {
+        workspace: Workspace,
+    },
+    GetWorkspaceSize {
+        workspace: Workspace,
+    },
```
mahkoh: Same here.

jay-config/src/_private/ipc.rs
```diff
@@ -1170,6 +1202,22 @@ pub enum Response {
...
+        height: i32,
+    },
+    GetWorkspacePosition {
+        x: i32,
+        y: i32,
+    },
+    GetWorkspaceSize {
+        width: i32,
+        height: i32,
+    },
```
mahkoh: Same here.

src/ifs/wl_surface/x_surface/xwindow.rs
```diff
@@ -308,9 +308,18 @@ impl Xwindow {
...
+                    .initial_floating_size(&self.toplevel_data)
+                    .unwrap_or_else(|| (ext.width(), ext.height()));
+                let abs_pos = self
+                    .data
+                    .state
+                    .initial_floating_position(&self.toplevel_data);
                 self.data
                     .state
-                    .map_floating(self.clone(), ext.width(), ext.height(), &ws, None);
+                    .map_floating(self.clone(), width, height, &ws, abs_pos);
```
mahkoh: Factor out `self.data.state`.

mahkoh:
I'd prefer for this to be a single action `set-position` that takes the following arguments:

- x1
- y1
- x2
- y2
- width
- height

All fields take numbers as arguments but also the string "keep", to indicate that this value should be preserved. Naturally not all combinations are valid. For example, it would not make sense to set x1=0, x2=100, width=keep, since you cannot follow all three instructions in general.

Ktrompfl:
Is this what you have in mind or should I keep the size / position setters as separate functions in jay-config and only expose them as a combined action in toml-config?

mahkoh:
I've only looked at the SetPosition data structure for now, not at the business logic, but I don't think it can be correct as is. The three states number/"keep"/omitted must be distinguished. Currently it seems that you conflate "keep" and omitted.

If a user sets x1=2, then I'd expect x2 to stay as is, so the window width might increase or decrease.

However, if a user sets x1=2,width="keep", then x2 should be adjusted dynamically to preserve the width.

Ktrompfl:
Alright, I've extended the set-position parameters to distinguish these cases. For ambiguous situations I defined the following behavior:
- If for an axis only one coordinate constraint is provided, the second coordinate is preserved and the size changes accordingly on this axis.
- If for an axis only a size constraint is provided, the first coordinate is preserved and the second coordinate is updated on this axis.
- If for an axis three constraints are provided but unsatisfiable, the changes to the other axis are not applied either and entire action is a no-op.

jay-config/src/input.rs
```diff
@@ -716,6 +717,21 @@ impl Seat {
...
+    ///
+    /// See [`Window::set_position`](crate::window::Window::set_position) for details.
+    pub fn set_position(
+        self,
+        x1: Option<Coordinate>,
+        y1: Option<Coordinate>,
+        x2: Option<Coordinate>,
+        y2: Option<Coordinate>,
+        width: Option<Coordinate>,
+        height: Option<Coordinate>,
```
mahkoh, line 730: I think this might be better as

```rust
#[derive(Default)]
struct Position {
    x1: Option<Coordinate>,
    ...
}

File: jay-config/src/video.rs [hunk 6]
```diff
@@ -201,6 +201,16 @@ impl Connector {
         get!().connector_size(self).1
     }
 
+    /// Returns the logical size of the connector.
+    ///
+    /// This is a shortcut for `(width(), height())` that only performs a single round-trip.
```
mahkoh, line 206: This is unnecessary.

jay-config/src/video.rs
```diff
@@ -201,6 +201,16 @@ impl Connector {
...
+    pub fn size(self) -> (i32, i32) {
+        if !self.exists() {
+            return (0, 0);
+        }
+        get!((0, 0)).connector_size(self)
```
mahkoh, line 211: Why not just `get!().connector_size(self)`?

jay-config/src/window.rs
```diff
@@ -66,6 +66,15 @@ pub enum TileState {
...
 }
 
+/// A constraint on a coordinate or size passed to [`Window::set_position`].
+#[derive(Serialize, Deserialize, Copy, Clone, Debug, Hash, Eq, PartialEq)]
+pub enum Coordinate {
+    /// Constrain this field to the window's current value for it.
+    Keep,
+    /// Constrain this field to this value.
+    Set(i32),
+}
```
mahkoh, line 76: I think this should be something like

```rust
enum Adjustment {
    #[default]
    Recalculate,
    Preserve,
    Set(i32),
}
```

mahkoh:
This PR does several independent things. Please split it up.

Ktrompfl:
I've split this into https://github.com/mahkoh/jay/pull/1158 and https://github.com/mahkoh/jay/pull/1159.

I don't really have a use case for the combined set-window-position action anymore and think its interface might be too complicated, so I'm leaving that out.

PR #1084, opened 2026-07-11
PoC: Native Sway-like Autotiling

GGchung69:
This is a proof of concept to demonstrate how native Sway-like autotiling (golden ratio splitting) can be implemented directly within Jay. 

### What it does:
- Introduces an `autotile` state flag to `WlSeat`.
- Updates the `toggle-split` configuration action to toggle this daemon on and off instead of manually cycling split directions.
- Hooks into window creation (`do_map_tiled`) to check the aspect ratio of the currently focused window. If `autotile` is enabled, it automatically wraps the focused window in a new split container along its longest axis before adding the new window.

This matches the exact behavior of external Sway autotiling scripts, providing dynamic nested containers instead of just changing the layout direction of the existing parent container.

*Note: Submitted as a PoC for discussion. Feedback is welcome!*


GGchung69:
Sorry, I post this with the help of gemini, slopping is maxing

jay-config/src/lib.rs
```diff
@@ -92,14 +92,16 @@ pub enum Direction {
 pub enum Axis {
     Horizontal,
     Vertical,
+    Auto,
```
mahkoh, line 95: Auto is not an axis.

jay-config/src/input.rs
```diff
@@ -429,7 +429,7 @@ impl Seat {
 
     /// Toggles the split axis of the parent-container of the currently focused window.
     pub fn toggle_split(self) {
-        self.set_split(self.split().other());
+        self.set_split(Axis::Auto);
```
mahkoh, line 432: This is obviously not acceptable as is.

src/ifs/wl_seat.rs
```diff
@@ -257,6 +257,7 @@ pub struct WlSeatGlobal {
     focus_history_rotate: NumCell<u64>,
     focus_history_visible_only: Cell<bool>,
     focus_history_same_workspace: Cell<bool>,
+    pub autotile: Cell<bool>,
```
mahkoh, line 260: I don't believe this should be a per-seat setting.

src/state.rs
```diff
@@ -1044,6 +1044,24 @@ impl State {
...
+                        let axis = if width > height {
+                            crate::tree::ContainerSplit::Horizontal
+                        } else {
+                            crate::tree::ContainerSplit::Vertical
+                        };
+                        crate::tree::toplevel_create_split(self, la, axis);
+                    }
+                }
+            }
+        }
```
mahkoh, line 1064: This logic probably belongs in map_tiled_on and you also need to handle the case where the parent is already a single-element container. You probably also don't want to use toplevel_create_split for this.

mahkoh (changes requested):
Having this kind of autotiling is ok as an option. However, this implementation is far from ok. You can probably do better by having the AI work harder.

GGchung69:
Thank you for trying to get this to become real code instead of just being some AI slop code, thank you so much

GGchung69:
tf does that button do?

GGchung69 (commented):
Better than my AI slop code

GGchung69:
Still have no idea what button do what

PR #1094, opened 2026-07-14
fix idle inhibition by decoupling from surface visibility toggles

shurizzle:
The old code synchronized inhibitors between per-surface and global maps on every visibility change via set_visible_() and detach_node_(). This was fragile: per-surface inhibitor entries disappeared during visibility toggles, leaving stale entries in the global map that could not be reactivated when the surface became visible again.

Refactor the design so inhibitors live in the global map from creation until explicit destruction or client disconnect. The idle task now evaluates surface visibility at check time rather than relying on activate/deactivate hooks on visibility changes.

- wl_surface: add is_visible() public accessor for SplitView<Cell<bool>>
- set_visible_(): replace inhibitor activate/deactivate loop with inhibitors_changed trigger so the idle task re-evaluates inhibition
- detach_node_(): remove inhibitor deactivate loop entirely
- install(): always call activate() regardless of surface visibility
- idle: check inhibitor surface visibility via is_visible() instead of testing only global map length
- state: make inhibitors_changed() public for cross-module access

### DIFF
--- a/src/ifs/wl_surface.rs
+++ b/src/ifs/wl_surface.rs
@@ -1807,6 +1807,10 @@ impl WlSurface {
         self.latch_listener.attach(&output.latch_event);
     }
 
+    pub fn is_visible(&self) -> bool {
+        self.visible[LiveTL].get()
+    }
+
     pub fn set_visible(self: &Rc<Self>, visible: bool) {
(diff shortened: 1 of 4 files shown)


mahkoh:
No thanks. If there is a bug with the code, open an issue.

PR #1113, opened 2026-07-19
wl_surface: center layer surface

### LINKED ISSUE #1092: Noctalia dock positioning issue (zwlr_layer_surface_v1 issue?)
Hello, just want to say I've really been enjoying using this compositor, especially for all the gaming features and protocols.

I've been using Jay with the noctalia shell but I've noticed that the dock anchors itself to an incorrect position (the bottom left instead of the bottom center like on other compositors supported by noctalia). I'm not a programmer so I had to ask Claude where the issue could potentially be, and it identified an issue with zwlr_layer_surface_v1.rs where there seems to be a missing/incorrect fallback to set a surface's position to the center if it's not assigned to the left or right.

I'm not sure if this is a correct fix but after building Jay with this patch the Noctalia dock does center itself correctly like it should, and other hud elements like the OSD popups and notifications are showing in the correct location as well

```
--- a/src/ifs/wl_surface/zwlr_layer_surface_v1.rs
+++ b/src/ifs/wl_surface/zwlr_layer_surface_v1.rs
@@ -541,6 +541,8 @@
             x1 = ml;
         } else if anchor.contains(RIGHT) {
             x1 = owidth - width - mr;
+        } else {
+            x1 = (owidth - width) / 2;
         }
         if anchor.contains(TOP | BOTTOM) {
             y1 = mt + (oheight - height - mt - mb) / 2;
@@ -548,6 +550,8 @@
             y1 = mt;
         } else if anchor.contains(BOTTOM) {
             y1 = oheight - height - mb;
+        } else {
+            y1 = (oheight - height) / 2;
         }
         let a_rect = Rect::new_sized_saturating(x1 + rect.x1(), y1 + rect.y1(), width, height);
         let o_rect = a_rect.move_(-opos.x1(), -opos.y1());
```

witchlliee:
fixes #1092

### DIFF
--- a/src/ifs/wl_surface/zwlr_layer_surface_v1.rs
+++ b/src/ifs/wl_surface/zwlr_layer_surface_v1.rs
@@ -537,8 +537,11 @@ impl ZwlrLayerSurfaceV1 {
         let extents = self.surface.extents.get();
         let (width, height) = extents.size();
         let mut anchor = self.anchor.get();
-        if anchor == 0 {
-            anchor = LEFT | RIGHT | TOP | BOTTOM;
+        if anchor & (LEFT | RIGHT) == 0 {
+            anchor |= LEFT | RIGHT;
+        }
+        if anchor & (TOP | BOTTOM) == 0 {
+            anchor |= TOP | BOTTOM;
         }
         let (mt, mr, mb, ml) = self.margin.get();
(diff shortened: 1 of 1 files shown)


mahkoh (approved):

mahkoh:
Thanks!

PR #1117, opened 2026-07-19
wl_surface: don't propagate commit to ext if ext has changed since commit

### LINKED ISSUE #1103: Incorrect popup_done() for client destroyed popups
XdgPopup.destroy_node() responds to client destroyed popups with popup_done. This should not happen. It should only be sent for compositor initiated dismissal.

mahkoh:
Closes #1103


witchlliee:
It does indeed fixes the issue. Thanks!

PR #1128, opened 2026-07-24
feat: per-window theme overrides

michaeladler:
Theme properties can now be overridden per window via window rules, e.g.:

```toml
[[windows]]
match.app-id = "mpv"
action = { type = "set-window-theme", theme.border-width = 0 }
```

If several rules override the same property of a window, the rule that appears last takes precedence.

### DIFF
--- a/jay-config/src/_private/ipc.rs
+++ b/jay-config/src/_private/ipc.rs
@@ -1084,6 +1084,16 @@ pub enum ClientMessage<'a> {
         target: ContainerTarget,
         axis: RelativeAxis,
     },
+    SetWindowSize {
+        window: Window,
+        sized: Resizable,
+        size: Option<i32>,
+    },
+    SetWindowColor {
+        window: Window,
+        colorable: Colorable,
+        color: Option<Color>,
+    },
 }
 
 #[derive(Serialize, Deserialize, Debug)]
--- a/jay-config/src/_private/ipc.rs
+++ b/jay-config/src/_private/ipc.rs
@@ -1018,6 +1018,16 @@ pub enum ClientMessage<'a> {
         reuse: bool,
     },
     GetSplitReusesContainer,
+    SetWindowSize {
+        window: Window,
+        sized: Resizable,
+        size: Option<i32>,
+    },
+    SetWindowColor {
+        window: Window,
+        colorable: Colorable,
+        color: Option<Color>,
+    },
 }
 
 #[derive(Serialize, Deserialize, Debug)]
--- a/src/theme.rs
+++ b/src/theme.rs
@@ -12,6 +12,7 @@ use jay_config::theme::BarPosition as ConfigBarPosition;
 use jay_config::theme::ContainerBorders as ConfigContainerBorders;
 use jay_proc::jay_clone;
 use linearize::Linearize;
+use linearize::StaticMap;
 use std::cell::Cell;
 use std::cmp::Ordering;
 use std::ops::Add;
@@ -638,6 +639,8 @@ pub struct Theme {
     pub show_window_icons: Cell<bool>,
     pub window_icons_grayscale: Cell<bool>,
     pub container_borders: SplitView<Cell<ContainerBorders>>,
+    /// An empty set of overrides used by views without per-window overrides.
+    empty_overrides: ThemeOverrides,
 }
 
 impl Default for Theme {
@@ -655,11 +658,25 @@ impl Default for Theme {
             show_window_icons: Cell::new(true),
             window_icons_grayscale: Cell::new(false),
             container_borders: Default::default(),
+            empty_overrides: Default::default(),
         }
     }
 }
 
 impl Theme {
+    /// Returns a view of this theme without any per-window overrides.
+    pub fn view(&self) -> ThemeView<'_> {
+        ThemeView::new(self, &self.empty_overrides)
+    }
+
+    /// Returns a view of this theme through `overrides`.
+    pub fn view_with_overrides<'a>(
+        &'a self,
+        overrides: Option<&'a ThemeOverrides>,
+    ) -> ThemeView<'a> {
+        ThemeView::new(self, overrides.unwrap_or(&self.empty_overrides))
+    }
+
     pub fn title_font(&self) -> Arc<String> {
         self.title_font.get().unwrap_or_else(|| self.font.get())
     }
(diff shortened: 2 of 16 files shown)


toml-spec/spec/spec.yaml
```diff
@@ -4444,6 +4444,16 @@ WindowRule:
         
         If a window matches any rule for which this is false, the window will not be
         automatically focused.
+    border-width:
```
mahkoh: Fields inside window rules themselves are only for things that must be known synchronously since actions in window rules execute asynchronously. This is not one of these cases.

This should instead be a regular action called "set-float-border-width" that takes a parameter specifying the width.

jay-config/src/_private/ipc.rs
```diff
@@ -748,6 +748,10 @@ pub enum ClientMessage<'a> {
         matcher: WindowMatcher,
         auto_focus: bool,
     },
+    SetWindowMatcherBorderWidth {
```

mahkoh:
~Sorry I was mistaken about this being a separate action. This size should be configured just like any other size. I.e. it should closely follow the existing border-width, bar-height, etc. settings. No new action is required for this.~

mahkoh:
Forget my previous comment. I had forgotten that this was supposed to be a per-window setting. In that case some more machinery might be required. We will probably want to build some general-purpose machinery that allows arbitrary theme settings to be overwritten on per-window basis. While this action would be ok for this particular situation, it would not do if in the future we want other things, like the border color, to be settable for each window.

michaeladler:
I like that idea. I've updated my branch to allow per-window theme overrides. Seems to work fine so far. Once we agree upon the design I will update the docs.

toml-spec/spec/spec.yaml
```diff
@@ -4448,6 +4448,98 @@ WindowRule:
...
+        Theme properties that are overridden for windows matching this rule.
+        
+        - Example:
+        
+          ```toml
+          [[windows]]
+          match.app-id = "org.kde.dolphin"
+          theme.border-width = 0
+          theme.focused-border-color = "#ff0000"
+          ```
```
mahkoh: Didn't we discuss that this should not be a new field in the window rule?

toml-spec/spec/spec.yaml
```diff
@@ -4448,6 +4448,98 @@ WindowRule:
...
+
+
+WindowTheme:
```
mahkoh: Is this necessary? Why not reuse the existing type?

toml-spec/spec/spec.yaml
michaeladler: To clarify: you mean to drop the `theme.` prefix? I.e. the above would become

```toml
[[windows]]
match.app-id = "org.kde.dolphin"
border-width = 0
focused-border-color = "#ff0000"
```

toml-spec/spec/spec.yaml
michaeladler: `WindowTheme` is a subset of `Theme` which contains the properties that "make sense" to define on a per-window basis, e.g. `bar_height` is part of `Theme` but doesn't make sense for a window, so it's not part of `WindowTheme`. We could use `Theme` of course and log a runtime warning instead, ignoring non-sensible properties.

toml-spec/spec/spec.yaml
mahkoh: No. This should be a regular action just like `set-theme`. Except that it acts on the per-window theme rather than the global theme.

michaeladler: Oh I see, I obviously misunderstood your last comment. I'll change it again to be an action, like we had in some previous version but applicable for the whole theme.

toml-spec/spec/spec.yaml
michaeladler: Switched to `Theme` now

michaeladler:
- Switched back to action-based window rule + applicable to `Theme`
- Rebased

src/theme.rs
```diff
@@ -669,8 +675,102 @@ impl Theme {
...
+/// A view of the theme through an optional set of overrides.
+///
+/// All code that renders or lays out a window should retrieve theme properties through a
+/// view instead of accessing [`Theme`] directly. That way per-window overrides apply
+/// automatically to every property.
+#[derive(Copy, Clone)]
+pub struct ThemeView<'a> {
+    pub theme: &'a Theme,
+    pub overrides: Option<ThemeOverrides>,
+}
```
mahkoh, line 763: Why not

```suggestion
#[derive(Copy, Clone)]
pub struct ThemeView<'a> {
    pub theme: &'a Theme,
    pub overrides: &'a ThemeOverrides,
}
```

?

mahkoh:
This looks pretty good!

src/theme.rs
michaeladler: `&'a ThemeOverrides` doesn't work here, mainly because the overrides live behind interior mutability. They're stored in `ToplevelData` in a `Cell`/`CloneCell`, which can only hand out values, never borrows. Getting a reference would mean switching to `RefCell` and carrying a `Ref` guard in `ThemeView`, which turns every layout/render call site into a potential runtime borrow panic - and the config handler does mutate overrides while views can be alive.

The lifetimes don't line up either: `'a` in `ThemeView<'a>` comes from the global `Theme`, while the overrides belong to a toplevel that can be dropped or overwritten at any time, and views are constructed from temporaries like `child.tl_data()`.

I've changed it to `Option<Rc<ThemeOverrides>>` though: the `Rc` gives the view a snapshot that stays valid even if a transaction commit or a config event replaces the toplevel's overrides underneath it.

michaeladler:
Successfully tested latest version

src/theme.rs
mahkoh: Why not just use

```rust
pub struct ThemeOverrides {
    pub sizes: StaticCopyMap<ThemeSized, Cell<Option<i32>>>,
    pub colors: StaticCopyMap<ThemeColored, Cell<Option<Color>>>,
}
```

?

michaeladler: Nice, the type tetris now works out. `ThemeView` is now `Copy`.

src/theme.rs
```diff
@@ -679,8 +696,100 @@ impl Theme {
...
+/// enough to add it to the `sizes!`/`colors!` macro invocations above.
+///
+/// Note that overrides only affect the parts of the tree that are owned by a single
+/// window. Sizes therefore only take effect for floating windows since the borders and
+/// title bars of tiled windows are shared between siblings.
+#[derive(Clone, Default, Debug, PartialEq)]
+pub struct ThemeOverrides {
+    pub sizes: StaticMap<ThemeSized, Cell<Option<i32>>>,
+    pub colors: StaticMap<ThemeColored, Cell<Option<Color>>>,
+}
```
mahkoh, line 742:
```suggestion
pub struct ThemeOverrides {
    pub sizes: StaticMap<ThemeSized, SplitView<Cell<Option<i32>>>>,
    pub colors: StaticMap<ThemeColored, SplitView<Cell<Option<Color>>>>,
}
```

src/tree/toplevel.rs
```diff
@@ -503,6 +505,8 @@ pub struct ToplevelData {
     pub session: CloneCell<Option<Rc<ToplevelSession>>>,
     pub is_root_container: SplitView<Cell<bool>>,
     pub is_overlay_root_container: Cell<bool>,
+    /// The theme overrides that apply to this toplevel, if any.
+    pub theme_overrides: SplitView<CloneCell<Option<Rc<ThemeOverrides>>>>,
```
mahkoh:
```suggestion
    pub theme_overrides: ThemeOverrides,
```

src/tree/toplevel.rs
```diff
@@ -411,6 +412,7 @@ pub enum ToplevelDataTransactionOp {
     SetWorkspace(Option<Rc<WorkspaceNode>>),
     SetVisible(bool),
     SetIsRootContainer(bool),
+    SetThemeOverrides(ThemeOverrides),
```
mahkoh: This shouldn't send the entire struct every time. There should be two enum variants here: one to update a single color and one to update a single size.

mahkoh:
This required tons of refactorings to fully implement it. The final result is now in #1263. Please test it.

PR #1132, opened 2026-07-25
container: configure solitary border width

### LINKED ISSUE #1091: Disable borders for solitary containers
With border colors enabled (https://github.com/mahkoh/jay/pull/1025), having one client open on a workspace keeps the focused border around it.
I find this behavior somewhat distracting, so I use the `smart_gaps` feature in Sway. It'd be nice to have something similar for Jay.

njdom24:
This allows me to specify `container-border-width-solitary = 0` for a "smart borders" setup to resolve https://github.com/mahkoh/jay/issues/1091.

### DIFF
--- a/src/tree/container.rs
+++ b/src/tree/container.rs
@@ -517,6 +517,10 @@ impl ContainerNode {
         }
     }
 
+    fn is_solitary(self: &Rc<Self>) -> bool {
+        self.num_children.get() <= 1 && self.toplevel_data.is_root_container.get()
+    }
+
     fn perform_layout(self: &Rc<Self>) {
         if self.num_children.get() == 0 {
             return;
@@ -549,7 +553,7 @@ impl ContainerNode {
 
         let theme = &self.state.theme;
         let th = theme.title_height(LiveTL);
-        let bw = theme.sizes.border_width.get(LiveTL);
+        let bw = theme.container_border_width(LiveTL, self.is_solitary());
         let num_children = self.num_children.get() as i32;
         let sp = match theme.container_borders[LiveTL].get() {
             ContainerBorders::Separators => 0,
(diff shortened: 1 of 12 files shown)


src/tree/container.rs
```diff
@@ -517,6 +517,16 @@ impl ContainerNode {
...
     }
 
+    fn is_solitary(self: &Rc<Self>, tl: TreeTimeline) -> bool {
+        if self.num_children.get() > 1 {
+            return false;
+        }
+        match self.workspace.get().node_state[tl].container.get() {
+            Some(root) => Rc::ptr_eq(&root, self),
+            None => false,
+        }
```
mahkoh: `self.toplevel_data.is_root_container`

src/tree/workspace.rs
```diff
@@ -248,11 +248,11 @@ impl WorkspaceNode {
             self.discard_child_properties(&*prev);
         }
         self.pull_child_properties(&**container);
+        self.set_ns_container(Some(container));
         let pos = ns.position.get();
         container.clone().tl_change_extents(&pos);
         container.tl_set_parent(self.clone());
         container.tl_set_visible(self.container_visible());
-        self.set_ns_container(Some(container));
```
mahkoh: What does this do?

mahkoh:
How would you know with this which window has the focus?

src/tree/workspace.rs
njdom24: I had some issues with an earlier iteration where it didn't pick up on the "solitary" state after I opened additional apps and moved them away. Will check it over.

njdom24:
> How would you know with this which window has the focus?

This isn't intended to only apply to focused containers. I can e.g. split it into focused/not if it's preferred.

mahkoh:
I don't know what you mean by that. Focused border colors were added to know which window has the focus. If you remove the border in this case, how do you know whether the window has the focus?

njdom24:
Good point. Smart borders / gaps is a feature I've used in Sway for a long time and carried over to others like Mango, so I haven't thought about it.
Generally, it works okay since my secondary monitors will have multiple apps and still have unfocused borders, or my solitary app will be Firefox, which fades out the tab bar when unfocused.
But it's true that having one app on each monitor, with this feature set to 0, otherwise makes focus ambiguous...

PR #1137, opened 2026-07-28
tree: make floating nodes containing only a placeholder invisible

### LINKED ISSUE #1134: Something is wrong with some games in fullscreen
Some games seems to also launch these weird floating windows when fullscreen, I can see them above the fullscreen if I enable float-above-fullscreen or sm: 

Image

[jay.log](https://github.com/user-attachments/files/30400651/jay.log)

[steam-418370.log](https://github.com/user-attachments/files/30400653/steam-418370.log)

Ktrompfl:
Fixes https://github.com/mahkoh/jay/issues/1134.

### DIFF
--- a/src/tree/float.rs
+++ b/src/tree/float.rs
@@ -93,8 +93,10 @@ pub struct FloatNode {
 #[derivative(Default)]
 pub struct FloatNodeState {
     pub visible: Cell<bool>,
+    pub requested_visible: Cell<bool>,
     pub position: Cell<Rect>,
     pub child: CloneCell<Option<Rc<dyn ToplevelNode>>>,
+    pub child_is_placeholder: Cell<bool>,
     #[derivative(Default(value = "Cell::new(WorkspaceType::Normal)"))]
     pub workspace_ty: Cell<WorkspaceType>,
     pub title_rect: Cell<Rect>,
@@ -210,7 +212,7 @@ impl FloatNode {
             cursors: Default::default(),
             transaction_data: TransactionData::new(&state.tree),
         });
-        floater.set_ns_visible(ws.float_visible());
+        floater.set_ns_requested_visible(ws.float_visible());
         floater.set_position(position);
         floater.set_ns_child(Some(&child));
         floater.set_ns_workspace_type(ws.ty);
@@ -224,12 +226,8 @@ impl FloatNode {
             .workspace_link
             .set(Some(ws.stacked.add_last(floater.clone())));
         child.tl_set_parent(floater.clone());
-        let ns = &floater.node_state[LiveTL];
-        child.tl_set_visible(ns.visible.get());
+        floater.update_effective_visible();
         child.tl_restack_popups();
-        if ns.visible.get() {
-            floater.display_link.borrow().invalidate();
-        }
         if child.tl_data().pinned.get() {
             floater.toggle_pinned();
         }
@@ -798,6 +796,11 @@ impl FloatNode {
         self.node_state[LiveTL].visible.replace(v)
     }
 
+    fn set_ns_requested_visible(self: &Rc<Self>, v: bool) {
+        self.add_transaction_op(FloatTransactionOp::SetRequestedVisible(v));
+        self.node_state[LiveTL].requested_visible.set(v);
+    }
+
     pub fn set_ns_position(self: &Rc<Self>, v: Rect) {
         self.add_transaction_op(FloatTransactionOp::SetPosition(v));
         self.node_state[LiveTL].position.set(v);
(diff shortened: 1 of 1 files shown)


mahkoh:
Please cache the node type and keep the externally set visibility separate from the effective visibility. While it might not be strictly necessary to store all three in the node state, that is what I'd prefer at the moment.

witchlliee:
No more placeholder with this PR. Thanks!

mahkoh:
What I had in mind was something like `fn update_effective_visible`. That is called from `new`, `set_visible`, and `replace_child` and combines the other two properties. Since the body would essentially be identical to the current `set_visible`, that function should be renamed and then a replacement added.

witchlliee:
still showing for minecraft on native wayland unfortunately

image

image

Ktrompfl:
This PR only affects placeholders for floating windows. Did you fullscreen the Minecraft NeoForge* 1.21.1 window from a tiled state and move it to another output or workspace? If so, the placeholder still showing is likely intended behavior: It reserves the window's position in the tiled grid so it can return there once it's no longer fullscreened.

witchlliee:
> This PR only affects placeholders for floating windows. Did you fullscreen the Minecraft NeoForge* 1.21.1 window from a tiled state and move it to another output or workspace? If so, the placeholder still showing is likely intended behavior: It reserves the window's position in the tiled grid so it can return there once it's no longer fullscreened.

Ahh got it. For some annoying reason glfw is launching the minecraft on the wrong monitor and the placeholder shows up but if I return it to my main monitor, there is no placeholder anymore.

Ktrompfl:
> Ahh got it. For some annoying reason glfw is launching the minecraft on the wrong monitor and the placeholder shows up but if I return it to my main monitor, there is no placeholder anymore.

The placeholder should still be there, just hidden below the fullscreen window as intended. Initially, it should open on the same monitor as the fullscreen window though, so maybe something goes wrong there. I haven't encountered this with Minecraft so far, maybe it only occurs on specific GLFW versions.

witchlliee:
> > Ahh got it. For some annoying reason glfw is launching the minecraft on the wrong monitor and the placeholder shows up but if I return it to my main monitor, there is no placeholder anymore.
> 
> The placeholder should still be there, just hidden below the fullscreen window as intended. Initially, it should open on the same monitor as the fullscreen window though, so maybe something goes wrong there. I haven't encountered this with Minecraft so far, maybe it only occurs on specific GLFW versions.

I didn't explained it quite right. Basically it launches on the right monitor but if I fullscreen it, it goes to the other monitor, then I see the placeholder where it was its last position in my primary monitor.

witchlliee:
> > > Ahh got it. For some annoying reason glfw is launching the minecraft on the wrong monitor and the placeholder shows up but if I return it to my main monitor, there is no placeholder anymore.
> > 
> > 
> > The placeholder should still be there, just hidden below the fullscreen window as intended. Initially, it should open on the same monitor as the fullscreen window though, so maybe something goes wrong there. I haven't encountered this with Minecraft so far, maybe it only occurs on specific GLFW versions.
> 
I didn't explained it quite right. Basically it launches on the right monitor but if I fullscreen it, it goes to the other monitor, then I see the placeholder where it was its last position in my primary monitor. It happens in sligthly "older" minecraft versions, yeah. Thankfully for 26.3, they replaced GLFW for SDL3, so better wayland support.

witchlliee:
~~Not sure what is happening here tho. The game doesn't actually gets fullscreen so I use the bind to fullscreen it and then this happens:~~


~~https://github.com/user-attachments/assets/fa0e21fd-fafe-4156-b649-4ce0399b91cf~~

~~Here is a log from the session if needed:~~
~~[jay.log](https://github.com/user-attachments/files/30511000/jay.log)~~

~~I guess this is the problem:~~

I got my Jay binary override smh and it wasn't making use of this PR, I built again and no placeholder anymore. The fullscreen problem still continues but I will move this to another issue.

src/tree/float.rs
```diff
@@ -774,14 +772,21 @@ impl FloatNode {
...
     }
 
     fn set_ns_child(self: &Rc<Self>, child: Option<&Rc<dyn ToplevelNode>>) {
         self.add_transaction_op(FloatTransactionOp::SetChild(child.cloned()));
-        self.node_state[LiveTL].child.set(child.cloned());
+        let ns = &self.node_state[LiveTL];
+        ns.child.set(child.cloned());
+        ns.child_is_placeholder
+            .set(child.is_some_and(|c| c.node_is_placeholder()));
     }
```
mahkoh, line 816: It's ok to set both fields in this function but the child_is_placeholder field must be transmitted through a separate FloatTransactionOp.

src/tree/float.rs
```diff
@@ -774,14 +772,21 @@ impl FloatNode {
         self.node_state[LiveTL].visible.replace(v)
     }
 
+    fn set_ns_requested_visible(self: &Rc<Self>, v: bool) {
+        self.node_state[LiveTL].requested_visible.set(v);
+    }
```
mahkoh, line 802: This needs to be transmitted through a FloatTransactionOp.

mahkoh (approved):

mahkoh:
Thanks!

PR #1147, opened 2026-08-01
ContainerBorders::FullSmart

### LINKED ISSUE #1091: Disable borders for solitary containers

khyperia:
Fixes https://github.com/mahkoh/jay/issues/1091

I only realized after writing this that somebody else already made a PR for this here https://github.com/mahkoh/jay/pull/1132 - but the implementation is decently different, so I'll just post this and let you choose which one to merge. I don't particularly mind either one!

### DIFF
--- a/src/theme.rs
+++ b/src/theme.rs
@@ -589,39 +589,50 @@ impl Into<ConfigBarPosition> for BarPosition {
 }
 
 #[derive(Copy, Clone, Debug, Hash, Eq, PartialEq, Default, Linearize)]
+pub enum ContainerBordersSetting {
+    #[default]
+    Separators,
+    Full,
+    FullSmart,
+}
+
+#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq, Default, Linearize)]
 pub enum ContainerBorders {
     #[default]
     Separators,
     Full,
 }
 
-impl StaticText for ContainerBorders {
+impl StaticText for ContainerBordersSetting {
     fn text(&self) -> &'static str {
         match self {
-            ContainerBorders::Separators => "Separators",
-            ContainerBorders::Full => "Full",
+            ContainerBordersSetting::Separators => "Separators",
+            ContainerBordersSetting::Full => "Full",
+            ContainerBordersSetting::FullSmart => "FullSmart",
         }
     }
 }
 
-impl TryFrom<ConfigContainerBorders> for ContainerBorders {
+impl TryFrom<ConfigContainerBorders> for ContainerBordersSetting {
     type Error = ();
 
     fn try_from(value: ConfigContainerBorders) -> Result<Self, Self::Error> {
         let v = match value {
-            ConfigContainerBorders::Separators => ContainerBorders::Separators,
-            ConfigContainerBorders::Full => ContainerBorders::Full,
+            ConfigContainerBorders::Separators => ContainerBordersSetting::Separators,
+            ConfigContainerBorders::Full => ContainerBordersSetting::Full,
+            ConfigContainerBorders::FullSmart => ContainerBordersSetting::FullSmart,
             _ => return Err(()),
         };
         Ok(v)
     }
 }
 
-impl Into<ConfigContainerBorders> for ContainerBorders {
+impl Into<ConfigContainerBorders> for ContainerBordersSetting {
     fn into(self) -> ConfigContainerBorders {
         match self {
-            ContainerBorders::Separators => ConfigContainerBorders::Separators,
-            ContainerBorders::Full => ConfigContainerBorders::Full,
+            ContainerBordersSetting::Separators => ConfigContainerBorders::Separators,
+            ContainerBordersSetting::Full => ConfigContainerBorders::Full,
+            ContainerBordersSetting::FullSmart => ConfigContainerBorders::FullSmart,
         }
     }
 }
@@ -637,7 +648,7 @@ pub struct Theme {
     pub bar_position: SplitView<Cell<BarPosition>>,
     pub show_window_icons: Cell<bool>,
     pub window_icons_grayscale: Cell<bool>,
-    pub container_borders: SplitView<Cell<ContainerBorders>>,
+    pub container_borders: SplitView<Cell<ContainerBordersSetting>>,
 }
 
 impl Default for Theme {
--- a/src/theme.rs
+++ b/src/theme.rs
@@ -589,39 +589,49 @@ impl Into<ConfigBarPosition> for BarPosition {
 }
 
 #[derive(Copy, Clone, Debug, Hash, Eq, PartialEq, Default, Linearize)]
-pub enum ContainerBorders {
+pub enum ContainerBordersSetting {
     #[default]
+    Separators,
+    Full,
+    FullSmart,
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum ContainerBorders {
     Separators,
     Full,
 }
 
-impl StaticText for ContainerBorders {
+impl StaticText for ContainerBordersSetting {
     fn text(&self) -> &'static str {
         match self {
-            ContainerBorders::Separators => "Separators",
-            ContainerBorders::Full => "Full",
+            ContainerBordersSetting::Separators => "Separators",
+            ContainerBordersSetting::Full => "Full",
+            ContainerBordersSetting::FullSmart => "FullSmart",
         }
     }
 }
 
-impl TryFrom<ConfigContainerBorders> for ContainerBorders {
+impl TryFrom<ConfigContainerBorders> for ContainerBordersSetting {
     type Error = ();
 
     fn try_from(value: ConfigContainerBorders) -> Result<Self, Self::Error> {
         let v = match value {
-            ConfigContainerBorders::Separators => ContainerBorders::Separators,
-            ConfigContainerBorders::Full => ContainerBorders::Full,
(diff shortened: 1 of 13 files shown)


njdom24:
I'm significantly less familiar with Jay's codebase (and Rust especially!). Your implementation looks cleaner and simpler to configure, so I'm happy to defer to it :D

njdom24:
This doesn't seem to work when I open a new window (e.g. `alacritty`) on an empty workspace. I need to open a second app, then close it (or move it to another workspace), for the now-single app to lose its border.

khyperia:
> This doesn't seem to work when I open a new window (e.g. `alacritty`) on an empty workspace. I need to open a second app, then close it (or move it to another workspace), for the now-single app to lose its border.

whoops! my bad, fixed now. please see [the force-push I just made](https://github.com/mahkoh/jay/compare/6684ae90197c40541d16087a85aaad3ad0452330..bd303a333f80f1aa7ff9c76ae61f125314d512f5): `tl_change_extents` now depends on the proper parent being set (specifically, `is_root_container`), so `tl_set_parent` must happen before it.

khyperia:
looks like we're both dealing with the same issue, in your PR as well there's some code mucking about that same function, presumably dealing with the same type of thing :P

src/tree/workspace.rs
```diff
@@ -255,8 +249,8 @@ impl WorkspaceNode {
         }
         self.pull_child_properties(&**container);
         let pos = ns.position.get();
-        container.clone().tl_change_extents(&pos);
         container.tl_set_parent(self.clone());
+        container.clone().tl_change_extents(&pos);
```
mahkoh, line 259: Why this change?

src/theme.rs
```diff
@@ -593,13 +593,22 @@ pub enum ContainerBorders {
...
     Full,
+    FullSmart,
+}
+
+#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq, Default, Linearize)]
+pub enum EffectiveContainerBorders {
+    #[default]
+    Separators,
+    Full,
 }
```
mahkoh: I prefer this to be called ContainerBorders and the other one ContainerBordersSetting or some such.

toml-spec/spec/spec.yaml
```diff
@@ -5115,6 +5115,8 @@ ContainerBorders:
       description: Only separators are drawn between children.
     - value: full
       description: A border is drawn around the entire container.
+    - value: full-smart
+      description: A border is drawn around the entire container, unless it is the only container on the workspace.
```
mahkoh: and has only one child.

src/tree/workspace.rs
mahkoh: Because of is_root_container I suppose.

mahkoh:
LGTM otherwise. I'd prefer to go with this.

khyperia:
done! thanks :heart:

src/tree/container.rs
```diff
@@ -1838,6 +1839,20 @@ impl ContainerNode {
         None
     }
 
+    fn container_borders(&self, timeline: TreeTimeline) -> ContainerBorders {
+        match self.state.theme.container_borders[timeline].get() {
+            ContainerBordersSetting::Separators => ContainerBorders::Separators,
+            ContainerBordersSetting::Full => ContainerBorders::Full,
+            ContainerBordersSetting::FullSmart => {
+                if self.num_children.get() == 1 && self.toplevel_data.is_root_container.get() {
```
mahkoh: These fields will need to be timelined as well.

khyperia: sorry for the delay, managed to do this now! did my best to figure out how to do this, not 100% confident though.

One question I had was whether to move `num_children` into `ContainerNodeState`. I've left it outside for now but I'm not particularly sure of the purpose/use of ContainerNodeState, if it's just "a bag of SplitView things" I should probably move it in, but if it's meaningful in some way it should stay outside.

mahkoh: In its current iteration the node_state fields exist primarily because most functions only deal with either live or render state and so we only need to index once.

So please put num children into the node state.

src/tree/container.rs
```diff
@@ -445,7 +446,8 @@ impl ContainerNode {
         new.tl_set_parent(self.clone());
         self.pull_child_properties(&new_ref);
         new.tl_set_visible(self.toplevel_data.visible[LiveTL].get());
-        let num_children = self.num_children.fetch_add(1) + 1;
+        let num_children = self.num_children[LiveTL].fetch_add(1) + 1;
+        self.add_transaction_op(ContainerTransactionOp::SetNumChildren(num_children));
```
mahkoh: Maybe a function like

```rust
fn adj_ns_num_children(&self, adj: impl FnOnce(usize) -> usize) -> usize;
```

src/tree/toplevel.rs
```diff
@@ -139,15 +139,20 @@ impl<T: ToplevelNodeBase> ToplevelNode for T {
...
-            data.is_root_container.set(true);
+            is_root_container = true;
             data.is_overlay_root_container
                 .set(ws.ty == WorkspaceType::Overlay);
         }
+        if data.is_root_container[LiveTL].replace(is_root_container) != is_root_container {
+            data.schedule_op(ToplevelDataTransactionOp::SetIsRootContainer(
+                is_root_container,
+            ));
+        }
```
mahkoh: This should be a function too to avoid the duplication below.

mahkoh:
LGTM otherwise.

src/tree/container.rs
khyperia: done! debated a bit between `adj: isize` with `usize::saturating_add_signed(isize)` and `adj: impl FnOnce(usize) -> usize`, but the fn is a bit clearer IMO, yeah.

src/tree/container.rs
```diff
@@ -175,6 +175,25 @@ pub struct ContainerNodeState {
...
+            mono_content: Default::default(),
+            abs_x1: Default::default(),
+            abs_y1: Default::default(),
+            width: Default::default(),
+            height: Default::default(),
+            content_width: Default::default(),
+            content_height: Default::default(),
+            num_children: NumCell::new(1),
+        }
+    }
```
mahkoh: I think I'd prefer not to have this impl and instead call adj_ns_num_children in Container::new just like with set_ns_split.

src/tree/container.rs
```diff
@@ -1231,10 +1249,6 @@ impl ContainerNode {
         self.update_title();
     }
 
-    pub fn num_children(&self) -> usize {
```
mahkoh, line 1234:
```rust
    pub fn num_children(&self, tl: TreeTimeline) -> usize {
```

mahkoh: Actually you can leave this as is.

mahkoh:
LGTM otherwise.

src/tree/container.rs
khyperia: done!

mahkoh:
Claude found two more issues:

- the new variant must be documented in the  book
- when is_root_container changes, we must trigger a layout of the container

For the second thing I'd suggest the following: Add a new function tl_is_root_container_changed and implement it in ContainerNode. Inline schedule_OP when changing this setting and additionally call the new callback.

Sent from my phone so the names might not be accurate.

khyperia:
> * when is_root_container changes, we must trigger a layout of the container
> 
> For the second thing I'd suggest the following: Add a new function tl_is_root_container_changed and implement it in ContainerNode. Inline schedule_OP when changing this setting and additionally call the new callback.

I'm not entirely sure what you mean, is this what you mean? (pushed)

src/tree/toplevel.rs
```diff
@@ -1092,6 +1096,16 @@ impl ToplevelData {
         }
     }
 
+    pub fn set_is_root_container(&self, value: bool) {
+        if self.is_root_container[LiveTL].replace(value) != value {
+            if let Some(slf) = self.slf.upgrade() {
```
mahkoh:
```suggestion
        if self.is_root_container[LiveTL].replace(value) != value &&
            let Some(slf) = self.slf.upgrade() {
```

src/tree/container.rs
```diff
@@ -2645,6 +2668,10 @@ impl ToplevelNodeBase for ContainerNode {
         }
     }
 
+    fn tl_is_root_container_changed(self: Rc<Self>) {
+        self.perform_layout();
```
mahkoh:
```suggestion
        self.schedule_layout();
```

Unless I missed something.

mahkoh:
> I'm not entirely sure what you mean, is this what you mean? (pushed)

Yes this looks good.

src/tree/container.rs
khyperia: right! that makes sense. pushed!

src/tree/container.rs
```diff
@@ -2645,6 +2670,11 @@ impl ToplevelNodeBase for ContainerNode {
         }
     }
 
+    fn tl_is_root_container_changed(self: Rc<Self>) {
+        self.update_content_size();
```
mahkoh, line 2674: Forgot about this.

mahkoh (approved):

mahkoh:
Thanks!

khyperia:
thanks so much! really appreciate it and all your work on jay :heart:

PR #1149, opened 2026-08-01
Add HDR screenshot option

### LINKED ISSUE #1146: HDR screenshots
`jay screenshot` seems to create screenshots that are tone-mapped to SDR. As far as I know there are very few ways to capture HDR screenshots on Linux through apps so this feature would be useful. Though I'm not sure how difficult this feature would be to implement.

hamburg3rz:
Most likely needs to be cleaned up a little but adds `jay screenshot --hdr10` as an option in jay. Closes #1146.

### DIFF
--- a/src/cli/screenshot.rs
+++ b/src/cli/screenshot.rs
@@ -6,8 +6,12 @@ use crate::cli::GlobalArgs;
 use crate::cli::ScreenshotArgs;
 use crate::cli::ScreenshotFormat;
 use crate::eventfd_cache::EventfdCache;
+use crate::format::Format;
 use crate::format::XRGB8888;
+use crate::format::XRGB2101010;
+use crate::format::formats;
 use crate::gfx_apis;
+use crate::ifs::jay_compositor::SCREENSHOT_DMABUF3_SINCE;
 use crate::tools::tool_client::Handle;
 use crate::tools::tool_client::ToolClient;
 use crate::tools::tool_client::with_tool_client;
@@ -25,8 +29,10 @@ use crate::video::drm::DrmError;
 use crate::video::gbm::GbmDevice;
 use crate::video::gbm::GbmError;
 use crate::wire::jay_compositor::TakeScreenshot;
+use crate::wire::jay_compositor::TakeScreenshot3;
 use crate::wire::jay_screenshot::Dmabuf;
 use crate::wire::jay_screenshot::Dmabuf2;
+use crate::wire::jay_screenshot::Dmabuf3;
 use crate::wire::jay_screenshot::DrmDev;
 use crate::wire::jay_screenshot::Error;
 use crate::wire::jay_screenshot::Plane;
@@ -36,13 +42,22 @@ use png::BitDepth;
 use png::ColorType;
 use png::Encoder;
 use png::SrgbRenderingIntent;
+use png::chunk;
 use std::cell::Cell;
 use std::cell::RefCell;
 use std::rc::Rc;
 use thiserror::Error;
 use uapi::OwnedFd;
 
+/// cICP chunk contents for HDR10: BT.2020 primaries, ST 2084 (PQ) transfer
+/// characteristics, identity matrix coefficients (PNG stores RGB, not YCbCr),
+/// full-range samples.
+const CICP_HDR10: [u8; 4] = [9, 16, 0, 1];
+
 pub fn main(_global: GlobalArgs, args: ScreenshotArgs) {
+    if args.hdr10 && args.format != ScreenshotFormat::Png {
+        fatal!("--hdr10 is only supported with --format=png");
+    }
     with_tool_client(|tc| async move {
         let screenshot = Rc::new(Screenshot {
             tc: tc.clone(),
@@ -61,10 +76,23 @@ async fn run(screenshot: Rc<Screenshot>) {
     let tc = &screenshot.tc;
     let comp = tc.jay_compositor().await;
     let sid = tc.id();
-    tc.send(TakeScreenshot {
-        self_id: comp,
-        id: sid,
-    });
+    let version = tc.jay_compositor_version().await;
+    if version >= SCREENSHOT_DMABUF3_SINCE {
+        tc.send(TakeScreenshot3 {
+            self_id: comp,
+            id: sid,
+            include_cursor: false,
+            hdr10: screenshot.args.hdr10,
+        });
+    } else {
+        if screenshot.args.hdr10 {
+            fatal!("The compositor does not support HDR10 screenshots");
+        }
+        tc.send(TakeScreenshot {
+            self_id: comp,
+            id: sid,
+        });
+    }
     let result = Rc::new(AsyncQueue::default());
     Error::handle(tc, sid, result.clone(), |res, err| {
         res.push(Err(err.msg.to_owned()));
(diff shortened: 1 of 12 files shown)


src/cli/screenshot.rs
```diff
@@ -61,9 +73,11 @@ async fn run(screenshot: Rc<Screenshot>) {
     let tc = &screenshot.tc;
     let comp = tc.jay_compositor().await;
     let sid = tc.id();
-    tc.send(TakeScreenshot {
+    tc.send(TakeScreenshot3 {
         self_id: comp,
         id: sid,
+        include_cursor: 0,
+        hdr10: screenshot.args.hdr10 as u32,
     });
```
mahkoh: This needs to check the jay_compositor version and then fall back to the older request if necessary.

src/cli/screenshot.rs
```diff
@@ -232,8 +266,38 @@ pub fn buf_to_bytes(
         ));
     }
 
+    let is_hdr10 = buf.format == XRGB2101010;
```
mahkoh: Please make this an explicit event instead of relying on this heuristic.

mahkoh: Actually forget this. Can't this simply be `screenshot.args.hdr10 && compositor supports this flag`?

src/screenshoter.rs
```diff
@@ -54,17 +58,33 @@ pub fn take_screenshot(
     if extents.is_empty() {
         return Err(ScreenshooterError::EmptyDisplay);
     }
+    // HDR10 output is 10-bit-per-channel BT.2020 primaries with an ST 2084 (PQ)
+    // transfer function. `windows_bt2100` is exactly that color description, so
+    // we reuse it here instead of introducing a separate one.
+    let format = if hdr10 { XRGB2101010 } else { XRGB8888 };
+    let (cd, blend_cd) = if hdr10 {
+        let cd = state.color_manager.windows_bt2100().clone();
```
mahkoh: This doesn't need to clone.

src/cli/screenshot.rs
mahkoh: Forget this whole comment chain. I'll create a new one.

src/cli/screenshot.rs
```diff
@@ -232,8 +266,38 @@ pub fn buf_to_bytes(
...
+
     let mut out = vec![];
-    {
+    if is_hdr10 {
```
mahkoh: This mixes two things:

1. If the format is XRGB2101010, then we need to do the decoding done in the branch.
2. If the color description is hdr10, then we need to set the cICP chunk.

These should be handled separately. So the `if` here should be guarded on `buf.format == XRGB2101010` directly. Then the writing of the chunk should be guarded on whether we requested hdr10 from the compositor.

The same chunk check should be added to the else branch.

The else branch should also be guarded on the format being XRGB8888 with an else branch added below that aborts with a fatal error saying that we don't know how to transform the format into PNG.

src/cli/screenshot.rs
```diff
@@ -232,8 +266,38 @@ pub fn buf_to_bytes(
...
+                image_data.extend_from_slice(&0xffffu16.to_be_bytes());
+            }
+        }
+        let mut encoder = Encoder::new(&mut out, buf.width as _, buf.height as _);
+        encoder.set_color(ColorType::Rgba);
+        encoder.set_depth(BitDepth::Sixteen);
+        let mut writer = encoder.write_header().unwrap();
+        // BT.2020 primaries, ST 2084 (PQ) transfer characteristics.
+        writer.write_chunk(chunk::cICP, &CICP_HDR10).unwrap();
+        writer.write_image_data(&image_data).unwrap();
```
mahkoh: Since this code is now going to be identical for both branches, the if/else should probably return the encoder and the image data, with this code moved out of the if/else.

src/ifs/jay_compositor.rs
```diff
@@ -53,6 +53,7 @@ use thiserror::Error;
 pub const CREATE_EI_SESSION_SINCE: Version = Version(5);
 pub const SCREENSHOT_SPLITUP_SINCE: Version = Version(6);
 pub const GET_TOPLEVEL_SINCE: Version = Version(12);
+pub const SCREENSHOT_HDR10_SINCE: Version = Version(39);
```
mahkoh: Rename this to SCREENSHOT_DMABUF3_SINCE.

mahkoh:
LGTM otherwise.

hamburg3rz:
Thank you for reviewing, I'll try my best to fix everything when I have time

hamburg3rz:
Pushed a new commit which should hopefully be an improvement. Apologies if not, I'm still quite new to git and rust.

src/cli/screenshot.rs
```diff
@@ -61,10 +75,24 @@ async fn run(screenshot: Rc<Screenshot>) {
...
+        });
+    } else {
+        if screenshot.args.hdr10 {
+            fatal!("The compositor does not support HDR10 screenshots");
+        }
+        tc.send(TakeScreenshot2 {
+            self_id: comp,
+            id: sid,
+            include_cursor: 0,
+        });
```
mahkoh: Why are you using TakeScreenshot2 here?

src/cli/screenshot.rs
```diff
@@ -122,7 +170,13 @@ async fn run(screenshot: Rc<Screenshot>) {
     };
     let format = screenshot.args.format;
     let eventfd_cache = EventfdCache::new(&tc.ring, &tc.eng);
-    let data = match buf_to_bytes(&eventfd_cache, drm_dev.as_ref(), &buf, format) {
+    let data = match buf_to_bytes(
+        &eventfd_cache,
+        drm_dev.as_ref(),
+        &buf,
+        format,
+        screenshot.args.hdr10,
```
mahkoh, line 181: This is not correct if the compositor does not support hdr10.

src/ifs/jay_compositor.rs
```diff
@@ -86,7 +87,7 @@ global_base!(JayCompositorGlobal, JayCompositor, JayCompositorError);
 
 impl Global for JayCompositorGlobal {
     fn version(&self) -> u32 {
-        38
+        39
```
mahkoh: Needs to be bumped to 40.

src/tools/tool_client.rs
```diff
@@ -350,7 +350,7 @@ impl ToolClient {
             self_id: s.registry,
             name: s.jay_compositor.0,
             interface: JayCompositor.name(),
-            version: s.jay_compositor.1.min(38),
+            version: s.jay_compositor.1.min(39),
```
mahkoh: Same here.

mahkoh:
LGTM otherwise.

wire/jay_compositor.txt
```diff
@@ -143,6 +143,12 @@ request create_keymap_builder (since = 32) {
     id: id(jay_keymap_builder) (new),
 }
 
+request take_screenshot3 (since = 39) {
```
mahkoh: Same here.

wire/jay_screenshot.txt
```diff
@@ -29,3 +29,10 @@ event dmabuf2 (since = 6) {
     height: i32,
     modifier: pod(u64),
 }
+
+event dmabuf3 (since = 39) {
```
mahkoh: Same here.

hamburg3rz:
I've tried to address the review comments so far. I'm not sure if we're fully there yet, but I'm testing this current build and it seems to work just fine.

src/cli/screenshot.rs
mahkoh: Sorry I was mistaken about this. I missed that you emit a fatal error if the compositor does not support hdr10. Please restore this as it was.

hamburg3rz: No problem, I'm pretty sure it's fixed now

mahkoh (approved):

mahkoh:
Thanks!

PR #1158, opened 2026-08-03
config: add getters for size and position

Ktrompfl:
Adds `size` and `position` methods for windows and workspaces and for consistency a `size` method for outputs (`position` already exists here). 

This is useful to create screenshots, e.g.
```rust
fn screenshot_workspace(seat: Seat) {
    let ws = seat.get_workspace();
    if !ws.exists() {
        return;
    }
    let (x, y) = ws.position();
    let (w, h) = ws.size();
    screenshot_region(x, y, w, h);
}
```

### DIFF
--- a/jay-config/src/_private/ipc.rs
+++ b/jay-config/src/_private/ipc.rs
@@ -1008,6 +1008,12 @@ pub enum ClientMessage<'a> {
     GetPlaneColorPipelinesEnabled {
         device: DrmDevice,
     },
+    GetWindowPosition {
+        window: Window,
+    },
+    GetWorkspacePosition {
+        workspace: Workspace,
+    },
 }
 
 #[derive(Serialize, Deserialize, Debug)]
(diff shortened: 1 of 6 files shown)


jay-config/src/_private/ipc.rs
```diff
@@ -1282,6 +1294,22 @@ pub enum Response {
...
+        height: i32,
+    },
+    GetWorkspacePosition {
+        x: i32,
+        y: i32,
+    },
+    GetWorkspaceSize {
+        width: i32,
+        height: i32,
+    },
```
mahkoh, line 1302: Please make these just one each that contains a `Rect`.

Ktrompfl: `jay-config` does not depend on `jay-algorithm`, so should I just add a new `Rect` struct to `jay-config`?

If I create combined `GetWindowPosition { pos: Rect, }` and `GetWorkspacePosition { pos: Rect, }`, should I also expose these with combined methods `position(self) -> Rect`? 

This would be inconsistent with the existing methods for connectors, which have  `position(self) -> (i32, i32)` and still require two messages to get both coordinates and size. 
Can and should there be a breaking change to make this consistent with windows and workspaces? 
If not, what would be a good name for the new methods to make this less confusing?

mahkoh: Just send the coordinates in a single message. You can keep the public functions as is. 

>Can and should there be a breaking change

There are no breaking changes in jay-config.

mahkoh (dismissed):

mahkoh (approved):

mahkoh:
Thanks!

Be aware that I changed the documentation somewhat.

PR #1159, opened 2026-08-03
config: add window rule to set initial floating size and position of floating windows

Ktrompfl:
Adds `initial-floating-size` and `initial-floating-position` window rules to solve https://github.com/mahkoh/jay/issues/1052.

### DIFF
--- a/src/config/handler.rs
+++ b/src/config/handler.rs
@@ -231,6 +231,20 @@ pub(super) struct ConfigProxyHandler {
             TileState,
         ),
     >,
+    pub window_matcher_initial_floating_size: CopyHashMap<
+        WindowMatcher,
+        (
+            Rc<CachedCriterion<WindowCriterionIpc, ToplevelData>>,
+            (i32, i32),
+        ),
+    >,
+    pub window_matcher_initial_floating_position: CopyHashMap<
+        WindowMatcher,
+        (
+            Rc<CachedCriterion<WindowCriterionIpc, ToplevelData>>,
+            (i32, i32),
+        ),
+    >,
 }
 
 pub struct ConfigWorkspace {
@@ -2675,6 +2689,9 @@ impl ConfigProxyHandler {
         self.window_matcher_leafs.remove(&matcher);
         self.window_matcher_no_auto_focus.remove(&matcher);
         self.window_matcher_initial_tile_state.remove(&matcher);
+        self.window_matcher_initial_floating_size.remove(&matcher);
+        self.window_matcher_initial_floating_position
+            .remove(&matcher);
     }
 
     fn handle_enable_window_matcher_events(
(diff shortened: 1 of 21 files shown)


src/config/handler.rs
```diff
@@ -231,6 +231,20 @@ pub(super) struct ConfigProxyHandler {
...
+            (i32, i32),
+        ),
+    >,
+    pub window_matcher_initial_floating_position: CopyHashMap<
+        WindowMatcher,
+        (
+            Rc<CachedCriterion<WindowCriterionIpc, ToplevelData>>,
+            (i32, i32),
+        ),
+    >,
```
mahkoh: I think it's time to replace all of these maps keyed on WindowMatcher by a single map that contains an `Rc` with `Cell` fields.

src/ifs/wl_surface/x_surface/xwindow.rs
```diff
@@ -327,11 +327,14 @@ impl Xwindow {
...
-                self.data
-                    .state
-                    .map_floating(self.clone(), ext.width(), ext.height(), &ws, None);
+                let state = &self.data.state;
+                let ws = state.float_map_ws();
+                let (width, height) = state
+                    .initial_floating_size(&self.toplevel_data)
+                    .unwrap_or_else(|| (ext.width(), ext.height()));
+                let abs_pos = state.initial_floating_position(&self.toplevel_data);
+                state.map_floating(self.clone(), width, height, &ws, abs_pos);
```
mahkoh: These calculations should not be done for each toplevel role individually.

src/ifs/wl_surface/xdg_surface/xdg_toplevel.rs
```diff
@@ -439,7 +439,11 @@ impl XdgToplevel {
     }
 
     fn map_floating(self: &Rc<Self>, workspace: &Rc<WorkspaceNode>, abs_pos: Option<(i32, i32)>) {
-        let (width, height) = self.toplevel_data.float_size(workspace);
+        let (width, height) = self
+            .state
+            .initial_floating_size(&self.toplevel_data)
+            .unwrap_or_else(|| self.toplevel_data.float_size(workspace));
+        let abs_pos = abs_pos.or_else(|| self.state.initial_floating_position(&self.toplevel_data));
```
mahkoh: I think this will also cause these settings to be applied when a window is moved between workspaces, which I don't think is correct.

Furthermore, when a window toggles between tiling and floating, they should not be applied either.

Ktrompfl:
I also changed the `initial-floating-position` rule to be output-relative, as this is more stable across output layout changes and is more intuitive when the floating node absolute position is [clamped](https://github.com/mahkoh/jay/blob/0f6dcc9101fe7690f70a149347fad86ccc3cc2a1/src/tree/float.rs#L186).

mahkoh (approved):

mahkoh:
Thanks!

PR #1175, opened 2026-08-11
Allow zwlr_foreign_toplevel_manager_v1 to move mouse to focused window

hamburg3rz:
I've been experimenting with getting Noctalia's window switcher to function properly with Jay. Some people want the mouse to follow the selected/focused window and I've found this small patch allows it to happen if `unstable-mouse-follows-focus = true` is in the jay config.


mahkoh (approved):

mahkoh:
Thanks!

Issue #1077, opened 2026-07-09
Implement a way to set mono mode as the default


danielfikko:
Well, currently, I'm configuring it this way:

```
[[clients]]
action = "toggle-mono"
```

The problem is that when I open at least 3 windows, mono mode turns off and the windows become vertically tiled.

EDIT: In reality, it seems that mono mode turns off when the number of open windows is odd. When the number is even, mono mode turns on!

I’m not sure if the way I’m currently configuring it is what you recommend for those who want to use mono mode as the default or if there’s a more appropriate way to implement it. If it’s the former, is this behavior when opening 3 windows intentional or is it some kind of issue? Is there a way to work around this?

Thank you.

danielfikko:
I don't know what the stacking mode is like on the i3, but maybe it's better than the mono/tabbed mode for people who like stacking WM (?)
https://i3wm.org/docs/userguide.html#_layout_mode_for_new_containers

mahkoh:
I think stacking in i3 is like mono in Jay (which is called default in i3, I think) except that the title bars appear below each other instead of next to each other. Jay doesn't support such a layout.

This feature should be very easy to implement. It's therefore a great first issue if someone wants to contribute to Jay. In terms of the actual feature, it should be no more than a 5 line change. The biggest part of the PR would be wiring everything up in the toml config, writing the toml spec, updating the book, and adding a toggle in the control center.

If nobody gets to it before me, then I'll get to it when I find the time.

danielfikko:
I see. No need to rush! I think the fact that you develop what you’re really into or enjoy is what makes jay so great.

Certainly, this might be a feature others would want, but honestly, I’ve already left jay and returned to labwc. As impressive as jay is (frame scheduling, SCHED_RR, competent implementation of all protocols), I just can’t get comfortable with a tiling WM.

Alphare:
> I think stacking in i3 is like mono in Jay (which is called default in i3, I think) except that the title bars appear below each other instead of next to each other. Jay doesn't support such a layout.

I have just switched to Jay (thanks for the work, it's great!) from Sway, and this is very much something I'm missing. The reason is that I have two vertical displays, and I like to stack certain windows in the workspaces assigned to those displays so that moving left or right is *always* moving outputs and not workspaces unless I explicitly want that.

Just thought I'd let you know that there is strong interest in a stacked mode :)

Ktrompfl:
You can now set mono mode as the default for new workspaces with window rules, e.g.
[...]

Issue #641, opened 2025-10-09
Option for autotiling


Stoppedpuma:
This is something that I feel Jay could offer that's missing from other "i3/sway-like" window managers. This can be done in sway/i3 with a script that interacts with the IPC, but this doesn't work in Jay and would be nice to include directly in Jay regardless. A great UX thing to have with this is container rotation, this basically would make a currently vertical container horizontal and vise versa, I feel Hyprland does rotation pretty well if I had to give another example of this in another software.

GGchung69:
it's been almost a year but no comment?

GGchung69:
Ok I was too impatience and do a thanos and asked gemini pro to do it myself and it work... should i upload the patch as a draft PR?
*It is AI slop code so very much not recommended to merge

GGchung69:
Might be useful as a Proof of Concept AI patch that I made, but 99% not recommend to merge any AI stuff, more bug might get introduce after all
Here's a video of how it work:

https://github.com/user-attachments/assets/91019ace-1ab7-4b1b-b808-0af5d4a4f305

GGchung69:
sorry for ai slop but i don't know how to make one myself but here: https://github.com/mahkoh/jay/pull/1084

PR #1176, opened 2026-08-11
config: add initial-layout option for workspaces

Ktrompfl:
Implements https://github.com/mahkoh/jay/issues/1077 per workspace.

Allows to configure the initial layout (`mono` / `tile`) and for tiled layout the split direction (`horizontal` / `vertical` / `major` / `minor`) per workspace.  
`major` splits along the workspace's longer axis and `minor` along its shorter one, based on the ratio of workspace width to height. This makes it possible to split dynamically vertically on vertical outputs without configuring `initial-output`.
If the initial layout or split direction is not configured, the current default of splitting horizontally is preserved. 

This option is designed extendable to account for future layout options, e.g. `dwindle` for auto tiling as proposed in https://github.com/mahkoh/jay/issues/641.

Finally, since `mono` appears to be the first table without fields, toml-spec is extended to account for this.

### DIFF
--- a/jay-config/src/_private/client.rs
+++ b/jay-config/src/_private/client.rs
@@ -82,6 +82,7 @@ use crate::window::WindowCriterion;
 use crate::window::WindowMatcher;
 use crate::window::WindowType;
 use crate::workspace::WorkspaceDisplayOrder;
+use crate::workspace::WorkspaceLayout;
 use crate::xwayland::XScalingMode;
 use bincode::Options;
 use futures_util::task::ArcWake;
@@ -2250,6 +2251,14 @@ impl ConfigClient {
             workspace,
             connector,
         });
+    }
+
+    pub fn set_workspace_initial_layout(
+        &self,
+        workspace: Workspace,
+        layout: Option<WorkspaceLayout>,
+    ) {
+        self.send(&ClientMessage::SetWorkspaceInitialLayout { workspace, layout });
     }
 
     pub fn parse_keymap_2(&self, builder: KeymapBuilder<'_>) -> Keymap {
--- a/jay-config/src/_private/ipc.rs
+++ b/jay-config/src/_private/ipc.rs
@@ -54,6 +54,7 @@ use crate::window::Window;
 use crate::window::WindowMatcher;
 use crate::window::WindowType;
 use crate::workspace::WorkspaceDisplayOrder;
+use crate::workspace::WorkspaceLayout;
 use crate::xwayland::XScalingMode;
 use serde::Deserialize;
 use serde::Serialize;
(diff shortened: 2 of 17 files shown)


mahkoh:
I feel like most of this should already be possible with window rules. Something like

```toml
[[windows]]
match.types = "container"
match.is-root-container = true
match.just-mapped = true
action = [ "show-all", "tile-major" ]
```

`is-root-container` and `tile-major` do not exist but could easily be added.

Ktrompfl:
Would you prefer the approach with window rules? In that case I'll close this PR and create a new PR to add `is-root-container` and `tile-major` / `tile-minor` instead.

mahkoh:
>Would you prefer the approach with window rules?

Yes. I always prefer using existing more general machinery over specialized machinery. For example, you said that

>This option is designed extendable to account for future layout options, e.g. dwindle for auto tiling as proposed in https://github.com/mahkoh/jay/issues/641.

But that is just a window rule that executes `tile-major` for every container.

Ktrompfl:
Alright, thanks. I've just seen that there is already a window match criteria to match workspaces. I really like how powerful window and client rules are getting together with actions.

Ktrompfl:
> ```toml
> [[windows]]
> match.types = "container"
> match.is-root-container = true
> match.just-mapped = true
> action = [ "show-all", "tile-major" ]
> ```

I don't think this can work right now:
 
The actions `tile-veritical` / `tile-horizontal` are documented in `toml-spec` as 

> Sets the split of the currently focused container to ...

However, in `jay-config` they call the `set_split` method, documented as

> Sets the split axis of the parent-container of the currently focused window.

Thereby, they always act on the parent container and the actions are a no-op when used on a root container. 

`show-all` / `show-single` have the same problem and to be consistent new actions like `tile-major` / `tile-minor` should follow this pattern. 

Changing the behavior in `jay-config` would be a breaking change, so the documentation of the `toml-config` should probably be clarified instead. It would be easy to add new actions, which operate on nodes directly if they are containers and their parent node otherwise. However, this would introduce quite a lot of new actions with unnecessarily complicated names to distinguish them from the existing actions.

mahkoh:
Maybe something like `{ type = "show-all", target = "self" }` could work.

Ktrompfl:
That sounds reasonable, then `target` could be either `self` or `parent` (default). I assume this still requires new methods in `jay-config` and additionally `toml-config` must allow parametrized actions with defaults for all parameters to be called as simple actions.

mahkoh:
>I assume this still requires new methods in jay-config

Yes.

>and additionally toml-config must allow parametrized actions with defaults for all parameters to be called as simple actions.

I don't think it needs to be that magic. Just add the names of the relevant simple actions to `ActionParser::parse_table`.

PR #1178, opened 2026-08-11
config: add is-root-container window match rule

Ktrompfl:
Allows to match the root container of a workspace in window rules, as discussed in https://github.com/mahkoh/jay/pull/1176.

### DIFF
--- a/src/criteria/tlm.rs
+++ b/src/criteria/tlm.rs
@@ -22,6 +22,7 @@ use crate::criteria::tlm::tlm_matchers::tlmm_client::TlmMatchClient;
 use crate::criteria::tlm::tlm_matchers::tlmm_content_type::TlmMatchContentType;
 use crate::criteria::tlm::tlm_matchers::tlmm_floating::TlmMatchFloating;
 use crate::criteria::tlm::tlm_matchers::tlmm_fullscreen::TlmMatchFullscreen;
+use crate::criteria::tlm::tlm_matchers::tlmm_is_workspace_container::TlmMatchIsWorkspaceContainer;
 use crate::criteria::tlm::tlm_matchers::tlmm_just_mapped::TlmMatchJustMapped;
 use crate::criteria::tlm::tlm_matchers::tlmm_kind::TlmMatchKind;
 use crate::criteria::tlm::tlm_matchers::tlmm_seat_focus::TlmMatchSeatFocus;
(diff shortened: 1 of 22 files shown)


src/criteria/tlm.rs
```diff
@@ -388,6 +395,10 @@ impl TlMatcherManager {
         self.just_mapped[true].clone()
     }
 
+    pub fn is_root_container(&self) -> Rc<TlmUpstreamNode> {
```
mahkoh: This also needs to be added to the control center and the book.

src/tree/toplevel.rs
```diff
@@ -148,6 +149,9 @@ impl<T: ToplevelNodeBase> ToplevelNode for T {
             data.is_overlay_root_container
                 .set(ws.ty == WorkspaceType::Overlay);
         }
+        if data.is_root_container.get() != was_root_container {
+            data.property_changed(TL_CHANGED_IS_ROOT_CONTAINER);
+        }
```
mahkoh: This is not the only place where this property changes. @khyperia has added a centralized function to set this property in #1147. You can copy that function into this PR.

mahkoh:
I'm not yet sure if calling this is-root-container is correct. Maybe it should be is-workspace-container.

src/criteria/tlm.rs
Ktrompfl: I'm not familiar enough with Rust to know if this is practical, but would it be feasible to eventually reduce the required changes to the config and its docs with code generation? Even a small change currently means separate updates across the TOML config + spec and book, the config crate and its comments, the control center, and sometimes the CLI. For example, it could be nice to also invoke actions from the CLI. However, since they are a part of the TOML config with their concrete parsers right now, this would require significant code duplication.

mahkoh: >I'm not familiar enough with Rust to know if this is practical, but would it be feasible to eventually reduce the required changes to the config and its docs with code generation?

I don't know.

>For example, it could be nice to also invoke actions from the CLI. However, since they are a part of the TOML config with their concrete parsers right now, this would require significant code duplication.

I believe that would best be handled by having the toml config create a separate socket.

wire/jay_window_match_builder.txt
```diff
@@ -76,3 +76,7 @@ request workspace {
 request content_types {
     v: pod(u64),
 }
+
+request is_workspace_container {
```
mahkoh: Needs a since attribute with a CLI and jay_compositor version bump.

mahkoh (approved):

mahkoh:
Thanks!

PR #1179, opened 2026-08-11
config: add target parameter to container actions & split-major / ... / tile-minor actions

Ktrompfl:
Adds a `target` parameter to all container actions to operate on the node itself (`self`), the parent node (`parent`) or the node itself if it is a container and its parent otherwise (`auto`); discussed in https://github.com/mahkoh/jay/pull/1176.

Adds `split-major` / `split-minor` / `tile-major` / `tile-minor` actions which derive their split / tile direction from the geometry, i.e. the ratio of targets width and height.

### DIFF
--- a/book/src/tiling.md
+++ b/book/src/tiling.md
@@ -31,6 +31,19 @@ The `tile-horizontal` action sets the container to horizontal, and
 `tile-vertical` sets it to vertical -- unlike `split-horizontal`/`split-vertical`
 which wrap the window in a new container first.
 
+`split-major` and `split-minor` derive the direction from the shape of the
+window instead: `split-major` splits along its longer side, `split-minor`
+along its shorter side. `tile-major` and `tile-minor` do the same for the
+container's shape. A square window counts as wide.
+
+```toml
+[shortcuts]
+alt-Return = ["split-major", { type = "exec", exec = "alacritty" }]
+```
+
+New terminals then appear next to a wide window and below a tall one, keeping
+tiles roughly square.
+
 If the window being split is the only window in its container, the
 [`split-reuses-container`](configuration/misc.md#split-reuses-container)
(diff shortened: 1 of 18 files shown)


mahkoh:
This looks good. I'll have to review it more carefully later.

src/config/handler.rs
```diff
@@ -2179,6 +2183,155 @@ impl ConfigProxyHandler {
...
+        &self,
+        seat: Seat,
+        target: ContainerTarget,
+    ) -> Result<(), CphError> {
+        let seat = self.get_seat(seat)?;
+        self.respond(Response::GetMono {
+            mono: seat.get_container_mono(target).unwrap_or(false),
+        });
+        Ok(())
+    }
```
mahkoh: Instead of adding these new functions, add a target parameter to the existing functions where appropriate.

toml-spec/spec/spec.yaml
```diff
@@ -1025,6 +1025,175 @@ Action:
               kind: number
               integer_only: true
```
mahkoh: There should be no empty line here.

book/src/tiling.md
```diff
@@ -102,6 +115,13 @@ alt-s = "show-single"   # Enter mono mode
 alt-a = "show-all"      # Exit mono mode
 ```
 
+## Target Container
+
+The split and mono actions change the parent container of the window they
+apply to. Their table form takes a `target` field to change that -- `"self"`
+for a focused container, `"auto"` for whichever of the two is a container.
+See [Container actions](configuration/shortcuts.md#container-actions).
```
mahkoh: The `split-` actions do not have a table form.

mahkoh (approved):

mahkoh:
Thanks!

PR #1190, opened 2026-08-15
Improve full-screen toggling behaviour

hamburg3rz:
This pull request adds 2 main things:
1. Allows switching to another window on the same workspace while a window is full-screen by temporarily exiting full-screen. For example, using a window switcher to switch to another window while full-screen will now allow that new window to be focused properly. Previously, this only worked when switching to a window on a different workspace.
2. Adds a new setting, `restore-fullscreen-on-reselect`, which defaults to `false`. The window that was taken out of full-screen is tracked, and when it is selected again — either by clicking on it or using a window switcher — it is automatically restored to its previous full-screen state. If full-screen is manually exited (by using a keybind) this does not apply.

The main aim of this is to make jay behave more like other compositors when dealing with full-screened windows. Please let me know if this can be improved in any way.


jay-config/src/_private/ipc.rs
```diff
@@ -610,6 +610,10 @@ pub enum ClientMessage<'a> {
         above: bool,
     },
     GetFloatAboveFullscreen,
+    SetRestoreFullscreenOnReselect {
+        enabled: bool,
+    },
+    GetRestoreFullscreenOnReselect,
```
mahkoh, line 616: What can I do to make people understand that new messages must be added at the bottom? I've thought about adding a comment at the top of the enum, but since the enum is so large, most people will probably not see a comment there.

jay-config/src/lib.rs
```diff
@@ -479,6 +479,18 @@ pub fn toggle_float_above_fullscreen() {
     set_float_above_fullscreen(!get_float_above_fullscreen())
 }
 
+/// Sets whether a temporarily unfullscreened window is restored to fullscreen when it is reselected.
```
mahkoh, line 482: "selected" is not a concept that exists in Jay. I don't know what it means.

src/ifs/wl_seat/event_handling.rs
```diff
@@ -1203,6 +1203,49 @@ impl WlSeatGlobal {
...
+        let node_id = toplevel.node_id();
+        if workspace.pending_refullscreen.get() != Some(node_id) {
+            return;
+        }
+
+        let seat = Rc::downgrade(self);
+        let state = self.state.clone();
+        let eng = state.eng.clone();
+        let yield_eng = eng.clone();
+        let future = eng.spawn("refullscreen", async move {
```
mahkoh, line 1224: I don't usually spawn new futures so liberally.

src/ifs/wl_seat/event_handling.rs
```diff
@@ -1203,6 +1203,49 @@ impl WlSeatGlobal {
...
+            yield_eng.yield_now().await;
```
mahkoh, line 1225: This is highly suspicious. Why is this here?

src/tree/workspace.rs
```diff
@@ -649,6 +660,12 @@ impl ContainingNode for WorkspaceNode {
     }
 
     fn cnode_make_visible(self: Rc<Self>, _child: &dyn Node) {
+        if let Some(fs) = self.node_state[LiveTL].fullscreen.get() {
+            fs.clone().tl_set_fullscreen(false, None);
```
mahkoh, line 664: This would be wrong if _child is the fullscreen toplevel.

mahkoh:
>Adds a new setting, restore-fullscreen-on-reselect, which defaults to false. The window that was taken out of full-screen is tracked, and when it is selected again — either by clicking on it or using a window switcher — it is automatically restored to its previous full-screen state. If full-screen is manually exited (by using a keybind) this does not apply.

I find this highly surprising. I don't know of any other compositor that behaves like this. KDE doesn't unfullscreen a window and then refullscreen it automatically.

hamburg3rz:
> I find this highly surprising. I don't know of any other compositor that behaves like this. KDE doesn't unfullscreen a window and then refullscreen it automatically.

I suppose I should have specified that it's mostly games that do this. If I'm in a fullscreen game in KDE or Niri for example, alt-tabbing to another window and then alt-tabbing back will return the game to a fullscreen state. Which doesn't happen currently on Jay. Though maybe I took the wrong approach here

mahkoh:
>If I'm in a fullscreen game in KDE or Niri for example, alt-tabbing to another window and then alt-tabbing back will return the game to a fullscreen state.

I don't know about Niri, but on KDE a window does not exit fullscreen when switching to another window.

hamburg3rz:
> I don't know about Niri, but on KDE a window does not exit fullscreen when switching to another window.

I suppose KDE isn't actually toggling fullscreen and it's just allowing any window to appear over a fullscreen window if you alt-tab. Maybe I'm just at odds with how a tiling window manager handles these things

mahkoh:
I've looked at niri and it seems it also doesn't unfullscreen windows when switching to a different window using the default alt-tab switcher.

hamburg3rz:
But niri allows you to escape that fullscreen window by alt-tabbing (on the same workspace) and then return to that fullscreened window by alt-tabbing back to it. That was my goal really.

jay-config/src/_private/ipc.rs
luvvlyjude: perhaps some contribution.md spec or smt could contain notes like this? including priorities like maintaining compatibility with existing config/workflows and so on?

Ktrompfl: Sorry, a comment at the top of the enum would definitely have helped me avoid this. There's also other aspects that would be great to document somewhere like where to add configuration (toml-config, jay-config, control center, cli, ...) or documentation (toml-spec, book, ...). 
I don't exactly know how to set it up, but e.g. [XLA](https://github.com/openxla/xla/pulls) has a banner above its PR tracker with a reference to guidelines. This could also be used to pin the issue template above the issue tracker.

mahkoh: https://github.com/mahkoh/jay/actions/runs/32274061236/job/96137137332?pr=1190

jay-config/src/lib.rs
hamburg3rz: I didn't want to fullscreen the window again whenever it was focused, because just mousing over it and having it fullscreen could be frustrating. I wanted it to be on a mouse-click or selecting it in a window switcher. I'm not sure what to call it in this case

mahkoh:
I see a potential design here. Mind you I have not thought too much about it, so even if you implement it I might ultimately reject it.

- Create a new toplevel node type, say PendingFullscreen.
- PendingFullscreen has as a child another ToplevelNode.
- PendingFullscreen is rendered by rendering the underlying ToplevelNode and on top an overlay that says something to the effect of "left click to re-fullscreen this node / right click to dismiss".
- When performing any click, the PendingFullscreen node is first replaced by the ToplevelNode. Then, if the click was a left click, the ToplevelNode is fullscreened.

What do you think?

hamburg3rz:
>     * PendingFullscreen is rendered by rendering the underlying ToplevelNode and on top an overlay that says something to the effect of "left click to re-fullscreen this node / right click to dismiss".
> 
>     * When performing any click, the PendingFullscreen node is first replaced by the ToplevelNode. Then, if the click was a left click, the ToplevelNode is fullscreened.
> 
I like the idea of right-clicking to dismiss. I just think the overlay could be a bit distracting or intrusive in some cases. If you were in a fullscreen browser for example, then you may still want to be able to see your browser window clearly when you exit fullscreen. Maybe the overlay could be a setting you could toggle

mahkoh:
Maybe the overlay could only appear when the mouse is over the window.

PR #1191, opened 2026-08-15
flake: skip broken tests + various improvements

Ktrompfl:
The recently added io_uring tests break the sandboxed package build in the nix flake again. To prevent this from happening over and over again, this disables all tests.

Moreover, this contains various other fixes and improvements for the flake:

- The NixOS and Home Manager modules now use the package exposed by the flake instead of the stable package from nixpkgs without the overlay.
- Optional runtime dependencies can now be disabled individually via `withOpenGL`, `withVulkan` and `withSqlite`, as long as one renderer remains.
- The overlay builds Jay against the package set of the consumer, and the Rust toolchain is obtained from `rust-overlay.lib.mkRustBin`, so no second nixpkgs is instantiated anywhere.
- Cargo dependencies are built in a separate derivation via crane, so changing any crate in the workspace no longer recompiles all dependencies.
- The package is built with the minimal Rust profile, which drops clippy, rustfmt and rust-docs from the build closure.
- `meta.license` is `gpl3Only`, matching `Cargo.toml`; `licenses.gpl3` is deprecated.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -42,14 +42,23 @@
           pkgconf,
           fontconfig,
           libgbm,
+          libglvnd,
           libinput,
           pango,
-          udev,
-          xkeyboard-config,
-          libglvnd,
           sqlite,
+          udev,
           vulkan-loader,
+          xkeyboard-config,
+
+          # Jay loads its renderers and the session database at runtime, so each of these can be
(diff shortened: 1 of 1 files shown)


mahkoh (approved):

mahkoh:
Thanks!

Ktrompfl:
That was fast, thanks!

PR #1209, opened 2026-08-23
config: add target to warp-mouse-to-focus and output mode to unstable-mouse-follows-focus

### LINKED ISSUE #1207: output based cursor warping
currently cursor warps to the last window on the output it is warped to. Is it possible to have an option to warp to the center of an output

Ktrompfl:
Closes https://github.com/mahkoh/jay/issues/1207.

Adds a `target` parameter (`output` / `window` / `workspace`) to the `warp-mouse-to-focus` action. The default remains `window` for backwards compatibility.

Adds multiple modes (`none` / `output` / `window`) to the `unstable-mouse-follows-focus` option. The mode `output` only warps when the focus changes to a different output. For backwards compatibility, boolean values are still supported with their current behavior (`false` is `none` / `true` is `window`). These modes match what is supported by the option `mouse_warping` (`output` / `container` / `none`) of sway.

I've considered also adding a `workspace` mode to the latter option, but I can't come up with reasonable semantics for it because workspace focus changes behave differently and e.g. can be triggered via the panel.


jay-config/src/input.rs
```diff
@@ -754,6 +768,30 @@ pub enum FocusFollowsMouseMode {
     False,
 }
 
+/// A mouse-follows-focus mode.
+#[derive(Serialize, Deserialize, Copy, Clone, Debug, Hash, Eq, PartialEq)]
+pub enum MouseFollowsFocusMode {
```
mahkoh, line 824:
```suggestion
#[derive(Serialize, Deserialize, Copy, Clone, Debug, Hash, Eq, PartialEq)]
#[non_exhaustive]
pub enum MouseFollowsFocusMode {
```

jay-config/src/input.rs
```diff
@@ -754,6 +768,30 @@ pub enum FocusFollowsMouseMode {
...
+/// The target that the cursor is warped to.
+#[derive(Serialize, Deserialize, Copy, Clone, Debug, Hash, Eq, PartialEq)]
+pub enum WarpTarget {
+    /// The focused window.
+    Window,
+    /// The workspace containing the keyboard focus.
+    Workspace,
+    /// The output containing the keyboard focus.
+    Output,
+}
```
mahkoh, line 847: Why does this have a Workspace variant but the other enum doesn't?

src/ifs/wl_seat.rs
```diff
@@ -2177,15 +2235,29 @@ pub async fn handle_warp_mouse_to_focus(state: Rc<State>) {
         state.pending_warp_mouse_to_focus.non_empty().await;
         state.eng.yield_now().await;
         while let Some(seat) = state.pending_warp_mouse_to_focus.try_pop() {
-            seat.warp_mouse_to_focus_scheduled.set(false);
+            let target = seat.warp_mouse_to_focus_scheduled.take();
```
mahkoh:
```suggestion
            let Some(target) = seat.warp_mouse_to_focus_scheduled.take() else { continue; };
```

jay-config/src/input.rs
Ktrompfl: > I've considered also adding a `workspace` mode to the latter option, but I can't come up with reasonable semantics for it because workspace focus changes behave differently and e.g. can be triggered via the panel.

Warping the cursor to the center of the workspace is very simple and can be useful to account for the offset added e.g. by the status bar.

However, automatically triggering these warps on focus changes across workspaces is much more complicated. There are many different ways to trigger a workspace change and I couldn't come up with reasonable and consistent semantics for when it should or should not warp, so this would need to hook into everything that eventually calls `show workspace` in some way to check what caused it, i.e. if it was triggered by a key bind, a pointer action, or something entirely else. 
Example: When a workspace is activated by clicking on the built-in status bar, it should certainly not warp the cursor. External status bars using the `ext-workspace` protocol should behave in the same way, but there is no way to tell why a client requests activation via this protocol and other clients could use it in a way which would justify warping.

If you think this mode should exist, I can look into implementing it again, or the set of supported modes can be extended at a later point.

Ktrompfl:
I've additionally added the mode `workspace` to the `unstable-mouse-follows-focus` option, which only warps when the focus changes to a different workspace. Notably, this mode does not warp the cursor when a workspace is activated via the `ext-workspace` protocol, as this is likely caused by a cursor action on an external status bar.

For backwards compatibility, the `window` mode keeps the existing behavior of warping the cursor to the center of the focused window on respective actions, even when the focused window did not actually change (e.g. when only one window exists on a workspace, already has focus, and the `show-workspace` is invoked on this workspace, the cursor is warped back to the center of the window). 
The `output` and `workspace` modes only warp the cursor to the center of the new output / workspace with focus when these actually changed. The behavior of `window` does not make sense here (why should we warp the cursor to the center of a workspace / output when refocusing the same window, but not when moving focus between two windows on the same workspace).

mahkoh:
This looks good. However, it mixes two things: the window mode always warps whereas the other modes only warp on change. Selecting the target seems to be an orthogonal concern. How about this:

```toml
[unstable-mouse-follows-focus]
warp-target = "center-of-output"
only-if-changed = true
```

jay-config/src/input.rs
```diff
@@ -804,6 +818,34 @@ pub enum FocusFollowsMouseMode {
...
+    /// focus changes.
+    Window,
+    /// The cursor is moved to the center of the focused workspace whenever the keyboard
+    /// focus moves to a different workspace.
+    Workspace,
+}
+
+/// The target that the cursor is warped to.
+#[derive(Serialize, Deserialize, Copy, Clone, Debug, Hash, Eq, PartialEq)]
+pub enum WarpTarget {
```
mahkoh: Make this non_exhaustive and apply the pattern from https://github.com/mahkoh/jay/pull/1270.

Ktrompfl:
> This looks good. However, it mixes two things: the window mode always warps whereas the other modes only warp on change. Selecting the target seems to be an orthogonal concern. 

They are not really orthogonal right now because the _target_ also determines the _trigger_, i.e. the respective option centers on windows/workspaces/outputs when _any_ keyboard action changes focus switches between windows/workspaces/outputs. 
I don't see any use in supporting all combinations of triggers and targets and it would just make the configuration more complicated.     
 
I also don't see any use case for warping the pointer to the center of a workspace/output when changing focus between windows within that workspace/output, so `only-if-changed` would really only make sense for the window target.     
It might be better to remove this behavior from the window option instead (as this never really was intended), but that would be a breaking change with the current behavior, which might be undesirable despite the option being marked unstable. The current behavior is useful to reset my pointer to the center when I lose track of it, but the same can just be achieved by manually binding one of the warping actions or even executing them with (a selection of) all focus actions.

mahkoh:
>It might be better to remove this behavior from the window option instead (as this never really was intended)

Let's do that then.

mahkoh:
And if we're already making a breaking change, also remove the true/false variants.

mahkoh:
I'm planning a new release relatively soon.

mahkoh:
I've written the release notes. Please also add release notes to this PR if you want to get into that release.

release-notes.md
```diff
@@ -396,6 +396,29 @@ change the split direction of the parent container.
 
 Thanks to @michaeladler for implementing this.
 
+### Mouse Follows Focus Modes
+
+The `unstable-mouse-follows-focus` setting now takes one of `none`, `output`,
+`workspace`, or `window`. The boolean values are no longer supported.
```
mahkoh, line 402: I had forgotten that this was about the unstable option. I had intentionally not included it in the release notes when it was originally added. But I guess since you've already written them we can keep it.

Ktrompfl: I can also remove it again, if you would prefer that.

Ktrompfl:
Moving the focused window / workspace also warps the pointer. To make it consistent with the behavior for focus changes, this now includes checks to only warp in this case when the focused window is actually moved across the respective target and not if the action is a no-op.

release-notes.md
mahkoh: No, let's leave it in.

mahkoh:
I'm not yet happy with this. I'd prefer to avoid having to capture the before value and then passing it back into maybe_schedule.

How about doing something like this instead: The handle_warp_mouse_to_focus handles both the maybe warping and the unconditional warping. Which it does is determined by a bit flag stored in the seat. Similarly, the data you current store in the `moved` parameter would instead be stored in a bit flag (possibly the same bit flag). The handle_warp_mouse_to_focus iteration that processes the seat would then consume those flags and act accordingly.

In the places where you currently calculate the before value, you would instead call a function that schedules the "maybe warping" part of handle_warp_mouse_to_focus unconditionally and calculates the before value if it has not already been calculated. The before value would be stored next to the bit flag for handle_warp_mouse_to_focus to consume when it handles the maybe warping case.

PR #1216, opened 2026-08-26
Add `stacked` mono style

Alphare:
Discoverability of the codebase and test writing was largely aided by AI to (majorly) speed up the process, but everything was reviewed/(re)written by me and I can vouch for the changes. Whether they're good enough is still up in the air, this is my first contribution. :)

This does the part which is interesting to me in #1077, I don't think I fully understand the OP's intent.


mahkoh:
This is good in principle. I've rebased it to test it and it seems to work.

One thing I noticed is that there are thick borders between the window titles. This border style is usually used for resizing which doesn't apply here. Maybe this should use the separator style.

jay-config/src/_private/ipc.rs
```diff
@@ -1084,6 +1085,24 @@ pub enum ClientMessage<'a> {
...
+        seat: Seat,
+        style: MonoStyle,
+    },
+    GetWindowMonoStyle {
+        window: Window,
+    },
+    SetWindowMonoStyle {
+        window: Window,
+        style: MonoStyle,
+    },
```
mahkoh, line 1105: All of these need a `ContainerTarget` parameter which was just added.

Alphare:
I've updated with a separate commit (I would have folded the changes in the stack but I don't use git often and it seems quite annoying to do). I've written the changes on a very slow laptop while on the road, so I didn't have a change to properly test them, hopefully it's still fine.

> One thing I noticed is that there are thick borders between the window titles. This border style is usually used for resizing which doesn't apply here. Maybe this should use the separator style.

Feel free to change this yourself since you will be able test the changes way more easily than I could at least for a couple of weeks. I remember the border being there but didn't know about the resizing feature, so I assumed it was a different style that didn't bother me.

mahkoh:
I've rebased your changes.

A few more things that I've noticed:

- Scrolling to switch between windows should probably work on the entire top area that contains the titles and not just on individual titles. Currently scrolling doesn't work when the cursor is over one of the borders between titles.
- move-left and move-right seem to act like move-up and move-down. But they should instead move the window out of the stacked container.

PR #1263, opened 2026-09-16
theme: add per-window theming

mahkoh:
This is not reviewed in any way. But I'm thoroughly burned out on this so I'm just putting it here for now.


michaeladler:
Thanks! I was testing this yesterday and didn't encounter any issues. Works for me :)

PR #1265, opened 2026-09-17
Restore window focus after resuming from idle

### LINKED ISSUE #1264: Focus the active window when resuming from idle
When jay returns from inactivity, no window is focused (as indicated by the gray title bar), requiring you to click on the window with the mouse to focus it again.

So, could jay automatically give focus to the window that was previously in focus?

I usually didn’t pay much attention to this until it happened while I was playing a game and my mouse cursor was invisible, which left me unable to move for a few seconds. It was a bit inconvenient.

yian04475-cmyk:
Closes #1264.

When resuming from idle, window visibility was restored but keyboard
focus remained on the root node. Restore the previous focus for each
seat after the compositor becomes visible again.

This also extends the existing idle integration test to verify that the
previously focused window receives keyboard focus after wake-up.

## Testing

- GitHub Actions `tests` workflow passed
- Build, unit tests, integration tests, musl check, and release check passed


mahkoh (approved):

mahkoh:
Thanks!

